PR #1, opened 2025-05-03
chore(backend): add linting workflow and initial project setup

carlobortolan:
- Setup linting via [flake8](https://github.com/pycqa/flake8) and formatting with [black](https://github.com/psf/black)
- Add initial backend structure under `backend/`
- Add CI/CD for linting (Black + Flake8)

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -1,0 +1,37 @@
+name: Backend Lint
+
+on:
+  push:
+    paths:
+      - "backend/**"
+  pull_request:
+    paths:
+      - "backend/**"
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.10, 3.11]
+    defaults:
+      run:
+        working-directory: ./backend
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "pip"
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      - name: Check formatting with Black
+        run: black --check .
+
+      - name: Run flake8 lint
+        run: flake8 .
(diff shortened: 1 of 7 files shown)


.github/workflows/python-lint.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    paths:
+      - "backend/**"
+  pull_request:
+    paths:
+      - "backend/**"
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
```
ricayanzon, line 13: This is probably not gonna change for the time of the project, but better pin a specific version of Ubuntu 
to avoid potential conflicts.

`runs-on: ubuntu-24.04`

backend/README.md
```diff
@@ -0,0 +1,16 @@
+# Backend
```
ricayanzon, line 1: Maybe it would be good to include here also how to start the server.

backend/requirements.txt
```diff
@@ -0,0 +1,2 @@
+black
```
ricayanzon, line 1: Pin versions of the packages used, just to avoid conflicts in case they update a package. Always painful to find bugs/incompatibilities if they come from updating the dependencies.

backend/README.md
carlobortolan: As there is no server yet, I left it intentionally blank, but I've now added a placeholder that we can update once we decide on a backend framework.

.github/workflows/python-lint.yml
carlobortolan: Thanks for pointing it out :+1: ; fixed in https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/pull/1/commits/629f95c5cd1e14d6c3f70687959d2170c73e121f.

backend/requirements.txt
carlobortolan: Fixed in https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/pull/1/commits/629f95c5cd1e14d6c3f70687959d2170c73e121f.

backend/README.md
ricayanzon: Just a placeholder is enough. Since you can run the "hello world" app with `python3 -m main`, I was thinking something like that. But any placeholder is fine for now.

PR #2, opened 2025-05-04
feat: initial readme

specialfish9:
Hey all,

With this PR I've structured the README file.  It is mainly a placeholder has the project hasn't fully kickstarted yet. Folder structure may be discussed together :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ containing curated, anonymized newspaper articles helping teams stay on top of
 relevant coverage.
 
 ## Project Structure
+```
 eutop-mediamind/
 ├── apps/                  # Main applications
 │   ├── frontend/          # Frontend app
@@ -26,6 +27,7 @@ eutop-mediamind/
 ├── .gitignore
 │
 └── README.md
+```
 
 ## Getting started
 TODO
--- a/README.md
+++ b/README.md
@@ -8,17 +8,13 @@ relevant coverage.
 ## Project Structure
 ```
 eutop-mediamind/
-├── apps/                  # Main applications
-│   ├── frontend/          # Frontend app
-│   └── backend/           # Backend app
+├── frontend/              # Frontend app
+│
+├── backend/               # Backend app
 │
 ├── database/              # DB migrations and configs
 │
-├── packages/              # Reusable libraries/modules (if any)
-│
 ├── infra/                 # Infrastructure
-│   ├── dev/
-│   └── prod/
 │
 ├── scripts/               # Project-level scripts
 │


README.md
```diff
@@ -1 +1,39 @@
...
+## Project Structure
+```
+eutop-mediamind/
+├── apps/                  # Main applications
+│   ├── frontend/          # Frontend app
+│   └── backend/           # Backend app
+│
+├── database/              # DB migrations and configs
+│
+├── packages/              # Reusable libraries/modules (if any)
```
ricayanzon, line 17: Do we need this? I think all packages will probably either be in the `frontend` or `backend` folder. I wouldn't start creating stuff which we might use, but don't do now.

README.md
```diff
@@ -1 +1,39 @@
...
+│
+├── infra/                 # Infrastructure
+│   ├── dev/
+│   └── prod/
```
ricayanzon, line 21: Not really a fan of this. It will probably only be one file per folder, I think we can keep the structure as simple as possible for now and adjust it once it grows.

README.md
```diff
@@ -1 +1,39 @@
...
+relevant coverage.
+
+## Project Structure
+```
+eutop-mediamind/
+├── apps/                  # Main applications
+│   ├── frontend/          # Frontend app
+│   └── backend/           # Backend app
+│
+├── database/              # DB migrations and configs
```
ricayanzon, line 15: This is part of the backend, I'd put this inside the backend folder (once we have it). For now, I think it can be left out completely tbh. We should wait until hooking up a database as long as possible, since it only makes things more complicated (i.e. migrations).

README.md
```diff
@@ -1 +1,39 @@
...
+
+MediaMind is a tool designed to keep client teams informed and prepared by 
+delivering daily press reviews. Each morning, it compiles and distributes PDFs
+containing curated, anonymized newspaper articles helping teams stay on top of
+relevant coverage.
+
+## Project Structure
+```
+eutop-mediamind/
+├── apps/                  # Main applications
```
ricayanzon, line 11: Maybe put `frontend` and `backend` in the root, since there's only gonna be one app?

README.md
specialfish9: Imho database stuff should be stored separately from BE. Also, I think it's important to set up the database in the early life of the project, as we probably have more time. If we postpone, the risk is to arrive close to a deadline with weird postgres errors.

README.md
specialfish9: We don't really know what the deployment will look like, but probably it will require more than one single file  per env I think. However, I'm okay with the 'keep it simple for the moment' part, so I got rid of the `dev` and `prod` folders

PR #3, opened 2025-05-05
added typescript linting

davidoof3141:
Added eslint to the frontend folder including github actions


frontend/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "license": "ISC",
+  "type": "commonjs",
+  "devDependencies": {
+    "@eslint/js": "^9.26.0",
+    "eslint": "^9.26.0",
+    "eslint-config-prettier": "^10.1.2",
+    "eslint-plugin-prettier": "^5.4.0",
+    "prettier": "^3.5.3",
+    "typescript": "^5.8.3",
+    "typescript-eslint": "^8.31.1"
```
ricayanzon, line 23: Pin the versions of the dependences (by removing the `^`). Everyone makes mistakes and bugs or incompatibilities introduced in newer versions of dependencies are super annoying to find. But by pinning the versions, we can update the dependencies when we want and check that they don't break anything.

frontend/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "lint:fix": "eslint . --ext .js,.jsx,.ts,.tsx --fix",
+    "prettier": "npx prettier --write ."
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "type": "commonjs",
```
ricayanzon, line 15:
```suggestion
  "type": "module",
```

Use module since it's newer and has many advantages over commonjs. Don't forget to also update the eslint.config.mjs accordingly.

ricayanzon:
Since you already installed TypeScript, it'd be nice if you could also create a `tsconfig.json` file. If you need help with it, let me know :)

PR #26, opened 2025-05-08
Database setup

davidoof3141:
Added credentials to https://www.notion.so/EUTOP-1d85f4484057805f9607e586fdcbcafc
And provided function for dependency injection in fastapi for supabase


backend/example.env
```diff
@@ -0,0 +1,3 @@
+# Database connection details
+SUPABASE_URL='https://your-url.supabase.co'
+SUPABASE_KEY='your-key'
```
ricayanzon, line 3: **never** commit a `.env` file, not even an `example.env`. actually, all `.env` files should be ignored by git. more junior devs could think it's okay to push `.env` files if they see this, or could think they have to modify this file and then push the updates. but I agree that this needs to be documented. add this to the readme file in a dedicated **Environment Variables** section. and also, can you please add `*.env` in `.gitignore`? it should not be possible to push any `.env` file by accident.

backend/src/db.py
```diff
@@ -0,0 +1,26 @@
+# db.py
```
ricayanzon, line 1: why this comment?

backend/src/db.py
```diff
@@ -0,0 +1,26 @@
...
+load_dotenv()
+
+SUPABASE_URL = os.getenv("SUPABASE_URL")
+SUPABASE_KEY = os.getenv("SUPABASE_KEY")
+
+
+def get_supabase_client() -> Client:
+    if not SUPABASE_URL or not SUPABASE_KEY:
+        raise Exception(
+            status_code=500, detail="Supabase credentials are not set"
```
ricayanzon, line 17:
```suggestion
            status_code=500, detail="Supabase credentials missing in environment variables"
```

A little more specific.

backend/README.md
```diff
@@ -10,19 +10,34 @@ Directory containing backend services and packages.
...
+    ```
+
 - Install dependencies:
 
   ```bash
   pip install -r requirements.txt
   ```
 
-- Start the server:
+- Run the example script:
```
ricayanzon, line 33: I'd leave this as **Start the server** section, since it might not change (the server might still be started by running the `main.py` file).

ricayanzon (changes requested):

backend/example.env
carlobortolan: Using an example env file to document required environment variables is actually a common and well-established convention (many opensource projects and even some frameworks (e.g., see https://github.com/laravel/laravel) include an example env as part of their scaffolding by default).
Putting env vars directly into the README can work for smaller projects, but becomes very cluttered once you have dozens of variables.
However, I agree that we should rename it to e.g., env.example (or similar) and gitignore all *.env files :+1:

backend/src/db.py
davidoof3141: I would prefer more generic error messages to not leak any configuration details

backend/src/db.py
davidoof3141: I asked copilot to rewrite my code and it generated it automatically.

backend/example.env
ricayanzon: sounds good, tbh I have never seen an `example.env` file. but sure, why not do it like that. thanks!

backend/src/db.py
ricayanzon: ok, convinced :)

ricayanzon (approved):

PR #45, opened 2025-05-14
chore(frontend): setup react

### LINKED ISSUE #42: Add React Application

milljoniaer:
Setups up the basic react app with shadcn as design library, some basic project structure and components are also included.

closes #42


frontend/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+COPY --from=development-dependencies-env /app/node_modules /app/node_modules
+WORKDIR /app
+RUN npm run build
+
+FROM node:20-alpine
+COPY ./package.json package-lock.json /app/
+COPY --from=production-dependencies-env /app/node_modules /app/node_modules
+COPY --from=build-env /app/build /app/build
+WORKDIR /app
+CMD ["npm", "run", "start"]
```
specialfish9, line 22: Shouldn't we use an http server (like nginx) to run the app in "production" mode?

frontend/app/app.css
```diff
@@ -0,0 +1,139 @@
...
+body {
+  @apply bg-white dark:bg-gray-950;
+
+  /*
+  Include if we want to have a dark mode 
+
+  @media (prefers-color-scheme: dark) {
+    color-scheme: dark;
+  }
+  */
```
ricayanzon, line 22: I would remove commented code. this is so easy to forget and then just remains there, polluting the codebase. if we should decide for a dark mode, we can always just quickly google and find exactly these lines

frontend/app/app.css
```diff
@@ -0,0 +1,139 @@
...
+  --chart-5: oklch(0.645 0.246 16.439);
+  --sidebar: oklch(0.208 0.042 265.755);
+  --sidebar-foreground: oklch(0.984 0.003 247.858);
+  --sidebar-primary: oklch(0.488 0.243 264.376);
+  --sidebar-primary-foreground: oklch(0.984 0.003 247.858);
+  --sidebar-accent: oklch(0.279 0.041 260.031);
+  --sidebar-accent-foreground: oklch(0.984 0.003 247.858);
+  --sidebar-border: oklch(1 0 0 / 10%);
+  --sidebar-ring: oklch(0.551 0.027 264.364);
+}
```
ricayanzon, line 130: remove if we don't have a dark mode

frontend/app/components/ui/button.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from "react";
+import { Slot } from "@radix-ui/react-slot";
+import { cva, type VariantProps } from "class-variance-authority";
+
+import { cn } from "~/lib/utils";
+
```
ricayanzon, line 6: maybe a short explanation as (JSDoc) on how to use the button would be nice? your decision

frontend/app/custom-components/text.tsx
```diff
@@ -0,0 +1,36 @@
...
+    // For first level headlines
+    const Tag = tag || "h1";
+    return (
+      <Tag className="text-4xl font-extrabold tracking-tight lg:text-5xl">
+        {children}
+      </Tag>
+    );
+  }
+
+  // TODO: add other headline levels
```
ricayanzon, line 32: mmmm not a fan of commiting TODOs. those remain there usually forever since no one feels responsible to do them and/or they just become outdated since code changes and don't make sense anymore. you explain above the component what it's for, that should be enough

frontend/app/pages/welcome/welcome.tsx
```diff
@@ -0,0 +1,75 @@
...
+        nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat.
+        Consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
+        labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et
+        accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no
+        sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
+        sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor
+        invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At
+        vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd
+        gubergren, no sea takimata
+      </p>
```
ricayanzon, line 63: I don't get it, isn't this also text? shouldn't this also be wrapped by a <Text> tag?

frontend/app/pages/welcome/welcome.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+      <Button asChild variant="link">
+        <span>
+          <SquareArrowOutUpRight />
+          <a href="https://www.eutop.com/de/" target="_blank">
+            Visit EUTOP
+          </a>
+        </span>
```
ricayanzon, line 71: mm what if you have links in texts? shouldn't that also be inside a <Text> tag?

frontend/package.json
```diff
@@ -1,24 +1,43 @@
...
+    "eslint": "^9.26.0",
+    "eslint-config-prettier": "^10.1.5",
+    "eslint-plugin-prettier": "^5.4.0",
+    "prettier": "^3.5.3",
+    "tailwindcss": "^4.1.4",
+    "tw-animate-css": "^1.2.9",
+    "typescript": "^5.8.3",
+    "typescript-eslint": "^8.32.1",
+    "vite": "^6.3.3",
+    "vite-tsconfig-paths": "^5.1.4"
```
ricayanzon, line 41: pin dependencies (by removing the `^`). super painful to find if a bug/incompatibility is introduced by a new version and you didn't manually update that version and have to find out why something is now not working, which worked earlier.

frontend/app/components/ui/button.tsx
milljoniaer: this is auto generated code by shadcn library, I will add a comment in the Readme

ricayanzon: yes, but shadcn also says they are not so much a library as a collection of components you can add and modify. I think it'd be nice to have them right in the component, since we will also probably adapt the components to match our styling and purposes. so I think it's also fine to add comments right inside the component. this way, the information also stays where a developer would naturally expect it. what do you think?

frontend/app/custom-components/text.tsx
milljoniaer: I normally agree, but in this case, in this case I considered it fine. But I can also remove it

frontend/app/pages/welcome/welcome.tsx
milljoniaer: Sorry, I dont get your remark here 😅

ricayanzon: sorry the comment was incomplete. I meant if links shouldn't also be inside a `<Text>` tag? Or, more generally, links should also use the Text component, or not?

milljoniaer: Hm, not sure about that, I have to think about that one, but since this is a POC and mainly for setup, lets leave it like that for now.

ricayanzon (approved):

PR #48, opened 2025-05-14
SSO supported authentication structure with Supabase Auth.

huaihaochuan:
Tested with sign-in, login, logout, and SSO with Google.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -1,2 +1,9 @@
 # Database connection details
-DATABASE_URL='postgresql://user:password@localhost:5432/mydatabase'
+DATABASE_URL='postgresql://user:password@localhost:5431/mediamind'
+
+# Supabase
+SUPABASE_URL=your-supabase-url
+SUPABASE_KEY=your-supabase-key
+SUPABASE_JWT_SECRET=your-supabase-secret-key
+# SUPABASE_REDIRECT_URL=https://your-domain.com/auth/callback
+SUPABASE_REDIRECT_URL=your-redirect-url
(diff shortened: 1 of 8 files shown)


carlobortolan (commented):
I haven’t looked at the code yet, but I had a quick clarification question to avoid confusion: I thought we agreed on Monday to switch to PostgreSQL instead of Supabase, since 1. internally they’re the same and  2. the customer prefers a self-hosted DB and is more familiar with PostgreSQL. 

Has this decision changed, or did I misunderstand something?

nikbauer720:
@carlobortolan you're right. Since both, backend and meta team startet with this, it might be a litte bit confusing. I didn't know about the change until you told me yesterday. I think the best is when somebody from the meta team keeps working on this branch an modifies it (if necessary; I also haven't looked at the code jet.)

huaihaochuan:
@carlobortolan I could continue integrating it using Keyloak;)

ricayanzon:
I'm not sure what to do here tbh. This is still incomplete if we change to Postgres without Supabase, since it misses hashing of pws, managing the JWT tokens (which ones are valid), and other things which fall away since we're not using Supabase anymore. maybe it makes more sense to close it altogether and use something like [clerk](https://clerk.com/), which has a free plan for basic auth and has the option to upgrade plan to integrate SSO. also seems like less work and we don't have to worry about maintaining security standards and infrastructure to store credentials safely. 

@huaihaochuan what's your opinion? would you rather adapt this branch to work without Supabase or close and use clerk instead?

huaihaochuan:
@ricayanzon Sure, I could do it, @davidoof3141 could have a look after PR.

PR #54, opened 2025-05-18
feat: pipeline to generate typescript types

specialfish9:
This pr adds an action to generate ts types starting from the open api definition from the be.


milljoniaer:
Nice, looks good to me, however, we can only test if it works, once a backend PR is setup, right?

And there seems to be a deployment error, can you please have a look at that one?

.github/workflows/generate-api-types.yml
```diff
@@ -0,0 +1,49 @@
...
+        node-version: 18
+
+    - name: Install openapi-typescript
+      run: |
+        npm install openapi-typescript typescript
+
+    - name: Generate Typescript types
+      run: |
+        mkdir -p frontend/api-types/
+        npx openapi-typescript /tmp/openapi.json -o ./frontend/api-types/v1.d.ts
```
milljoniaer, line 41:
```suggestion
        npx openapi-typescript /tmp/openapi.json -o ./frontend/types/api-types-v1.d.ts
```

ricayanzon:
can be merged from my side once the comments are resolved :)

frontend/types/api.ts
```diff
@@ -0,0 +1,25 @@
+import createClient from "openapi-fetch";
+import { createQueryHook } from "swr-openapi";
+import type { paths } from "./api-types-v1";
+
+const client = createClient<paths>({
+  baseUrl: "https://myapi.dev/v1/", // TODO change with proper be url
```
ricayanzon, line 6: what URL is expected here? do we not have that already?

frontend/types/api.ts
```diff
@@ -0,0 +1,25 @@
...
+    "/api/v1/users"
+  );
+
+  if (isLoading || !data) return "Loading...";
+
+  if (error) return `An error occured: ${error.message}`;
+
+  return <div>{ data.title } </div>;
+}
+*/
```
ricayanzon, line 25: what is this example for? I don't quite get it. but just from my feeling, maybe something that belongs to the readme?

frontend/package.json
```diff
@@ -22,6 +22,7 @@
     "react": "19.1.0",
     "react-dom": "19.1.0",
     "react-router": "7.5.3",
+    "swr-openapi": "^5.3.0",
```
ricayanzon, line 25:
```suggestion
    "swr-openapi": "5.3.0",
```

pin dependency, this allows for minor versions to be updated (e.g. `5.4.0`)

frontend/types/api.ts
milljoniaer: I think that is just an example for the frontend to see how this is supposed to work, correct? @specialfish9 

But such hooks for the api should not belong into the `types` subfolder, right? So rather /frontend/api/example.ts or something like that?

milljoniaer:
Just a general question, since this is a rather not too obvious PR and functionality, do you think we should add a readme somewhere? Since it is addressing both frontend & backend the root ReadMe might be a good place?

frontend/types/api.ts
specialfish9: Yes sure, I left it for you. You can move it wherever you prefer :D

I can also remove it from the PR @ricayanzon  @milljoniaer

specialfish9:
> Just a general question, since this is a rather not too obvious PR and functionality, do you think we should add a readme somewhere? Since it is addressing both frontend & backend the root ReadMe might be a good place?

Not sure if it's needed. At the end of the day, it's an automatic process. If you want, I can do that. Documentation never killed anyone afaik 😆

ricayanzon:
> > Just a general question, since this is a rather not too obvious PR and functionality, do you think we should add a readme somewhere? Since it is addressing both frontend & backend the root ReadMe might be a good place?
> 
> 
> 
> Not sure if it's needed. At the end of the day, it's an automatic process. If you want, I can do that. Documentation never killed anyone afaik 😆 

I think it'd be nice to add it to the readme just for informing people that it's there and they don't need to create the types manually. would be nice if you could add it 🙂

frontend/types/api.ts
milljoniaer: Its fine for me, we have to adjust it anyways, so fine for me, you can leave it for this PR 👍

ricayanzon: mm yeah not really a fan of "we will do that later, leave it there" because then it gets forgotten and remains there, until someone asks what that is for and no one remembers. if this is an example, just put it into the readme, that's always a save bet. not such a fan of `example`-files either, because then we move away from it and I promise you, the least thing you will think about is to update the `example.ts`. the readme at least is, or should be, the single starting point, so it's easier to remember to update that or see that something is outdated.

ricayanzon (approved):

PR #55, opened 2025-05-18
feat/clerk-authentication-backend

huaihaochuan:
The setup is able to:
-Create user
-List all users (with the token from the frontend)


backend/app/api/v1/endpoints/user.py
```diff
@@ -1,8 +1,35 @@
...
+    return {
+        "id": current_user["id"],
+        "email": (
+            current_user["email_addresses"][0]["email_address"]
+            if current_user.get("email_addresses")
+            else None
+        ),
+        "first_name": current_user.get("first_name"),
+        "last_name": current_user.get("last_name"),
+    }
```
ricayanzon, line 35: this is the same endpoint as in `auth.py`, right? why? do we even need this endpoint in `auth.py`?

backend/app/api/v1/endpoints/auth.py
```diff
@@ -0,0 +1,94 @@
...
+                    "last_name": user.last_name,
+                }
+                for user in users
+            ]
+        }
+    except Exception as e:
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=f"Failed to fetch users: {str(e)}",
+        )
```
ricayanzon, line 57: doesn't seem like an endpoint that belongs to authentication. why is this here?

backend/app/api/v1/endpoints/auth.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@router.post("/signup")
+async def signup(user_data: UserCreate):
+    """
+    Sign up a new user
+    """
+    try:
+        with Clerk(bearer_auth=configs.CLERK_SECRET_KEY) as clerk:
```
ricayanzon, line 66: you're initializing the clerk client in the beginning of the file. maybe it makes more sense to remove it from the top and initialize it in every endpoint using `with`?

backend/app/api/v1/endpoints/user.py
```diff
@@ -1,8 +1,35 @@
-from fastapi import APIRouter
```
ricayanzon, line 1: maybe this file's name should be `users` instead of `user` and you move here the endpoint to get a list of all users. what do you think?

backend/app/api/v1/endpoints/user.py
```diff
@@ -1,8 +1,35 @@
...
+from pydantic import BaseModel
+
+from app.core.dependencies import get_current_user
+
+router = APIRouter(prefix="/users", tags=["users"])
 
 
 @router.get("", response_model=str)
 def get_user_list():
     return "hello world"
```
ricayanzon, line 13: isn't this implemented in `auth.py`? either remove this or move the endpoint for users here. please look out for stuff like this and maintain the codebase clean

backend/app/core/config.py
```diff
@@ -21,5 +21,9 @@ class Configs:
     # database
     DATABASE_URL = os.getenv("DATABASE_URL")
 
+    # Clerk Configuration
```
ricayanzon, line 24:
```suggestion
    # Configuration of the user management tool (Clerk)
```

backend/app/core/dependencies.py
```diff
@@ -0,0 +1,35 @@
...
+                )
+
+            user = response.json()
+            return user
+
+    except Exception:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Invalid authentication credentials",
+        )
```
ricayanzon, line 35:
```suggestion
            response.raise_for_status()
            return response.json()
            
    except httpx.HTTPStatusError as e:
        raise HTTPException(
            status_code=status.HTTP_401_UNAUTHORIZED,
            detail="Invalid authentication credentials",
        )
        
    except httpx.RequestError as e:
        raise HTTPException(
            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
            detail="Authentication service unavailable",
        )
```

duplicate exception raising

backend/app/core/dependencies.py
```diff
@@ -0,0 +1,35 @@
+import httpx
```
ricayanzon, line 1: btw, @nikbauer720 do we have logging for the backend?

backend/app/main.py
```diff
@@ -28,6 +29,7 @@ def __init__(self):
             )
 
         self.app.include_router(v1_routers, prefix="/api/v1")
+        self.app.include_router(auth_router, prefix="/api/v1/auth")
```
ricayanzon, line 32: you included the `auth_router` already in `routes.py`. this shouldn't be needed and only duplicates the routes

backend/app/api/v1/endpoints/auth.py
```diff
@@ -0,0 +1,94 @@
+from clerk_backend_api import Clerk
```
ricayanzon, line 1: I think this to be useful it should at least also implement login and logout functionality. can you please add that too? then the frontend team can connect it when they're done with the signup/login page

backend/app/core/dependencies.py
LeonardoZambranoL: Not yet. Issue is still open.

backend/app/api/v1/endpoints/auth.py
ricayanzon: solved, since that's not needed when using Clerk

backend/app/api/v1/endpoints/users.py
```diff
@@ -0,0 +1,59 @@
...
+            current_user["email_addresses"][0]["email_address"]
+            if current_user.get("email_addresses")
+            else None
+        ),
+        "first_name": current_user.get("first_name"),
+        "last_name": current_user.get("last_name"),
+    }
+
+
+@router.get("/users")
```
ricayanzon, line 35:
```suggestion
@router.get("")
```

`/users/users` is bad. remove this and move endpoint before the `/me` one.

backend/app/api/v1/endpoints/users.py
```diff
@@ -0,0 +1,59 @@
...
+                            if user.email_addresses
+                            else None
+                        ),
+                        "first_name": user.first_name,
+                        "last_name": user.last_name,
+                    }
+                    for user in users
+                ]
+            }
+    except Exception as e:
```
ricayanzon, line 55:
```suggestion
            }

    except Exception as e:
```

add empty line between first exception and try

backend/app/api/v1/endpoints/users.py
```diff
@@ -0,0 +1,59 @@
...
+        raise HTTPException(
```
ricayanzon, line 56: you're returning the same error regardless of the exception that is actually thrown. you don't have to put all here, but maybe some like `Unauthorized` would be good

nikbauer720 (approved):

PR #58, opened 2025-05-19
WIP - feat(logging): add centralized structured logger to backend

NicolasArteaga:
- Created app/core/logger.py to set up a reusable logger.
- Integrated the logger in main.py to log app startup.
- Tested it by running the FastAPI app and accessing the API docs.
- Log messages now show time, log level, and module name.
- Supports INFO, DEBUG, and ERROR levels for development and production use.
- Logger messages confirm application startup in the terminal.
- Updated requirements.txt with logger dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,17 @@ eutop-mediamind/
 └── README.md
 ```
 
-## Getting started
-TODO
+## Monitoring & Error Reporting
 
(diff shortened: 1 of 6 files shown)


LeonardoZambranoL:
Please dont push the venv.

PR #59, opened 2025-05-19
feat(logger): integrate structured logger and initialize in main

NicolasArteaga:
- Created app/core/logger.py to set up a reusable logger.
- Integrated the logger in main.py to log app startup.
- Tested it by running the FastAPI app
- Log messages show time, log level, and module name.
- Supports INFO, DEBUG, and ERROR levels for development and production use.
- Logger messages confirm application startup in the terminal.

### DIFF
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -1,0 +1,14 @@
+version: "3.8"
+
+services:
+  main:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: main
+    ports:
+      - "8000:8000"
+    restart: always
+    volumes:
+      - .:/app
+      - ./logs:/app/logs  # persistent logs directory
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.8"
 
 services:
   main:
@@ -12,3 +11,8 @@ services:
     volumes:
       - .:/app
       - ./logs:/app/logs  # persistent logs directory
+    logging:
+      driver: json-file
+      options:
+        max-size: "50m"
+        max-file: "10" #Max 10 files of 50MB each
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -178,4 +178,4 @@ cython_debug/
 .DS_Store
 
 # For Testing Scrapers:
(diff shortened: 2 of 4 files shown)


nikbauer720 (dismissed):

nikbauer720:
> maybe it would make sense to make the logs persistent outside of the docker container. what do you think?

Yes this would really makes sense as it gives us the opportunity to track why the demo environment fails.

NicolasArteaga:
I'm on it!

NicolasArteaga:
Perfect, I tested it and it worked alright! :)

NicolasArteaga:
Logger Enhancement (logger.py)
	•	Added a FileHandler to log messages to logs/app.log
	•	Ensures logs are persisted outside the container via a Docker volume

Docker Volume
	•	Volume mount: ./logs:/app/logs for persistent log storage

backend/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.8"
+
```
ricayanzon, line 2: versioning is obsolete (see [link](https://docs.docker.com/reference/compose-file/version-and-name/)), pls remove it :)

backend/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: main
+    ports:
+      - "8000:8000"
+    restart: always
+    volumes:
+      - .:/app
+      - ./logs:/app/logs  # persistent logs directory
```
ricayanzon, line 14: you're not specifying any `max-size` or other log configuration here, so theoretically, the logs could grow indefinitely. this could end up filling the entire disk space of the server. I think some basic options should be set here

NicolasArteaga: logging:
      driver: json-file
      options:
        max-size: "50m"
        max-file: "10" #Max 10 files of 50MB each
        
        I made it so that it rotates between files when they get to 50MB each and it has max 10 files, so 500MB of log memory which should be more than enough :)

NicolasArteaga: changed the logger.py too, to rotate the logs/file in the same style

nikbauer720 (approved):

PR #66, opened 2025-05-21
feat: added User Analytics using Hotjar

davidoof3141:
Frontend:
- root.tsx: simply added hotjar script tag to track user actions

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -4,3 +4,5 @@
 /.react-router/
 /build/
 .vercel
+
+.env
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -5,4 +5,4 @@
 /build/
 .vercel
 
-.env
+*.env
--- a/frontend/app/root.tsx
+++ b/frontend/app/root.tsx
@@ -32,6 +32,19 @@ export function Layout({ children }: { children: React.ReactNode }) {
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <Meta />
         <Links />
+        <script
+          dangerouslySetInnerHTML={{
+            __html: `(function (c, s, q, u, a, r, e) {
+              c.hj=c.hj||function(){(c.hj.q=c.hj.q||[]).push(arguments)};
+              c._hjSettings = { hjid: a };
+              r = s.getElementsByTagName('head')[0];
+              e = s.createElement('script');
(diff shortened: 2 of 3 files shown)


davidoof3141:
image
I linked hotjar to the preview url:
https://eutop-med-git-853bcc-center-for-software-engineering-excellence.vercel.app/
to check if the installation was successful. Once merged and deployed I have to manually change the url

frontend/app/root.tsx
```diff
@@ -55,6 +55,20 @@ export function Layout({ children }: { children: React.ReactNode }) {
...
+            __html: `(function (c, s, q, u, a, r, e) {
+              c.hj=c.hj||function(){(c.hj.q=c.hj.q||[]).push(arguments)};
+              c._hjSettings = { hjid: a };
+              r = s.getElementsByTagName('head')[0];
+              e = s.createElement('script');
+              e.async = true;
+              e.src = q + c._hjSettings.hjid + u;
+              r.appendChild(e);
+          })(window, document, 'https://static.hj.contentsquare.net/c/csq-', '.js', 6408235);`,
+          }}
```
milljoniaer, line 70: Not sure if I completely understood the code correctly, but why dont we just directly add the constructed script tag into head? If that works it would clean the code imo.

```html
<script src="https://static.hj.contentsquare.net/c/csq-6408235.js /> 
```

and I would create a static variable for the magic number `6408235`. I think adding it through environment variables might be over the top for now, but also something we could consider.

What do you think?

milljoniaer: I just tried to read what the code you took from the docs is actually doing, and that is basically adding the script tag, I came up with to the DOM. 

But I think I forgot that also some setting variable is added to the window object, so nevermind, lets do it, like it is done in the docs 👍

davidoof3141: It is done

ricayanzon (dismissed):

frontend/.gitignore
```diff
@@ -4,3 +4,5 @@
 /.react-router/
 /build/
 .vercel
+
+.env
```
ricayanzon, line 8:
```suggestion
*.env
```

ignore all files ending by .env, e.g. if someone has `local.env` or `prod.env` those should also be ignored.

ricayanzon (approved):

PR #68, opened 2025-05-21
feat: frontend pipelines

specialfish9:


.github/workflows/frontend-premerge.yml
```diff
@@ -1,21 +1,20 @@
...
   pull_request:
     paths:
       - "frontend/**"
+  workflow_call:
 
 jobs:
-  eslint:
-    runs-on: ubuntu-24.04
+  premerge:
+    runs-on: ubuntu-22.04
```
milljoniaer, line 11: just out of curiousity, why did you reduce the linux version?

milljoniaer (commented):
Looks good to me, but maybe we should wait for ricardos go?

.github/workflows/frontend-premerge.yml
specialfish9: Just because [act](https://github.com/nektos/act) doesn't support ubuntu 24.04 yet. I need act to test the pipelines locally, otherwise I would have to open trillions of PRs just to see if my last change works. 
However, usually I switch from 24 to 22 and then reset it to 24 before opening the PR, but this time I forgot about it lol

.github/workflows/frontend-deploy.yml
```diff
@@ -0,0 +1,29 @@
...
+  #   with:
+  #     image_name: ${{ DOCKER_IMAGE }}
+  #     image_tag: pr-${{ github.event.pull_request.number || github.run_number }}
+  #     dockerfile_path: ${{ DOCKERFILE_PATH }}
+  #     context: ${{ DOCKER_CONTEXT }}
+  #
+  #   secrets:
+  #     registry_username: ${{ secrets.DOCKER_USERNAME }}
+  #     registry_password: ${{ secrets.DOCKER_PASSWORD }}
+  #
```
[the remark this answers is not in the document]
specialfish9: I will, I swear :D

frontend/package-lock.json
ricayanzon, line 1: @carlobortolan just because I saw this now in another PR as well: the lock file is created automatically and should never be adjusted manually. it's created/updated when running `npm i` (or `npm ci` for "clean install" = will remove `node_modules` and install everything fresh). if there's still an unpinned version in this file but not the `package.json`, just run `npm i` and commit that.

carlobortolan: @ricayanzon Yes I know ;) 
If you're referring to [#82](https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/pull/82) I ran `npm i` and it automatically updated the package-lock.json (I didn't do that manually :sweat_smile:)

ricayanzon: ah ok, I was just confused why you suddenly had updates in `package-lock.json` but not in `package.json`. then someone forgot to do that previously.

ricayanzon (dismissed):

frontend/package.json
```diff
@@ -18,6 +22,7 @@
     "class-variance-authority": "0.7.1",
     "clsx": "2.1.1",
     "isbot": "5.1.27",
+    "jest": "29.7.0",
```
ricayanzon, line 25: oh please also move jest to devDependencies

milljoniaer (commented):
Do we want to add the frontend deployment in this PR already?

specialfish9:
> Do we want to add the frontend deployment in this PR already?

@ricayanzon correct me if I'm wrong, but we decided to keep the frontend on Vercel, so the deployment part should be already in place

ricayanzon (approved):

PR #74, opened 2025-05-22
Feat: added Qdrant vector database

davidoof3141:
**NEW**
*Scripts:*
- Added docker-compose.dev.yml for qdrant service

*Backend:*
- Added Qdrant client to core modules
- added qdrant url to env template

**FIX**
*Scripts:*
- fixed permission error of logs directory by using a named volume instead


scripts/docker-compose.dev.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.8'
+
```
ricayanzon, line 2: versioning is obsolete (see [link](https://docs.docker.com/reference/compose-file/version-and-name/)), pls remove it :)

also please see #59 since it will is also adding a `docker-compose.yml` and will probably be merged before this

backend/app/core/db.py
```diff
@@ -12,4 +18,15 @@ def get_postgresql_connection(cfg: Configs) -> PgConnection:
     except OperationalError as e:
         raise Exception(f"Failed to connect to PostgreSQL database: {str(e)}")
     except Exception as e:
+        logger.error(f"Failed to initialize PostgreSQL client: {str(e)}")
         raise Exception(f"Failed to initialize PostgreSQL client: {str(e)}")
+
+
+def get_qdrant_connection(cfg: Configs) -> QdrantClient:
+    if not cfg.QDRANT_URL:
+        raise ValueError("QDRANT_URL not set in config.")
```
ricayanzon, line 27: did you not log this on purpose or just forgot? 😄

PR #115, opened 2025-06-04
Fix/endpoints for user and search profile, topics and keywords

nikbauer720:
fixed the auth.py that it returns the internal user. 
Added endpoints for topics and keywords


carlobortolan (changes requested):
@nikbauer720 @niklasbauer20 
Why did you delete the terraform workflows, part of the backend-deploy workflows and the entire tf infra directory?

(and on a sidenote: why are you using two different git users?)

PR #152, opened 2025-06-07
fix: admin settings changes

MagMagnita:
- Added Breadcumbs and Fixed Header size
- Added Leonardo's suggestions
- Searchbar instead of filter
- New destructive dialogs
- Dialog to prevent deletion by mistake
- Password hiding / opening
- Track changes to remind user of unsaved ones when closing


frontend/.env.example
milljoniaer, line 1: why did you delete the `.env.example`?

Please revert this change

frontend/app/components/ui/button.tsx
```diff
@@ -19,6 +19,7 @@ const buttonVariants = cva(
           "bg-secondary text-secondary-foreground shadow-xs hover:bg-secondary/80",
         ghost:
           "hover:bg-accent hover:text-accent-foreground dark:hover:bg-accent/50",
+        ghostNoHover: "",
```
milljoniaer, line 22: why is this needed?

frontend/api/api-types-v1.ts
milljoniaer, line 1: I see why you added this, but please revert this in this PR as it is not related.

Also, we have the same types generated in `/types/api-types-v1.ts`, so if you generate types, please save them there

frontend/api/api.ts
milljoniaer, line 1: also unrelated, please revert

milljoniaer (commented):
Nice, thanks for your work, please have a look at these points:

1. can you please check, why the layout has a different width on the dashboard than on the admin page
2. can you please use the same design for the add user method as rafael does here
yours: 
image
raphael in edit profile:
image

milljoniaer (approved):
Looks good, thanks

PR #155, opened 2025-06-08
Feat/keyword generation

specialfish9:


backend/app/services/search_profiles_service.py
```diff
@@ -111,14 +113,21 @@ async def update_match_feedback(
...
-        # Avoid useless LLM calls if no profiles are available
-        if len(visible_search_profiles) == 0:
-            return []
+        class test:
+            def __init__(self, topic, keywords: List[str]):
+                kw = list(map(lambda x: Keyword(name=x), keywords))
+                self.topics = [Topic(search_profile_id=UUID(), name=topic, keywords=kw)]
 
-        lhm_client = LLMClient(LLMModels.openai_4o_mini)
+        visible_search_profiles = [test("trains", ["railway", "locomotive", "carriage", "track", "station"])]
```
nikbauer720, line 130: so it is mocked rn and so LLM call?

ricayanzon:
wtf happened here? 108 commits and >20k changes?? please let me know if this is a git problem and we should look into this together

backend/app/services/search_profiles_service.py
```diff
@@ -256,3 +263,27 @@ async def set_search_profile_subscriptions(
...
         )
+
+    @staticmethod
+    async def get_keyword_suggestions(user: User, suggestions: List[str]) -> KeywordSuggestionResponse:
+        # Avoid useless LLM calls if no topics are available
+        if len(suggestions) == 0:
+            return KeywordSuggestionResponse(suggestions=[])
+
+        prompt = """
+        I will give you a list related keywords. Please add 5 new relevant keywords.\n
```
milljoniaer, line 274: Just an idea, not sure if it makes sense.
```suggestion
        I will give you a list related keywords. Please add 5 new relevant keywords. Don't include synonyms, but suggest words to pin down the topic more exactly with your added relevant keywords. \n
```

frontend/app/custom-components/profile/edit/ai-suggestion-tag.tsx
```diff
@@ -19,8 +25,9 @@ export function AiSuggestionTag({ keyword }: AiSuggestionTagProps) {
       </Label>
       <div
```
milljoniaer, line 26: can we make it a button, so that screen readers no that its clickable?

frontend/app/custom-components/profile/edit/topics.tsx
```diff
@@ -26,9 +29,74 @@ export function Topics({ profile, setProfile }: TopicsProps) {
...
+
+  const keywordsString = selectedTopicKeywords.join(",");
+
+  useEffect(() => {
+    if (selectedTopic && selectedTopicKeywords.length > 0) {
+      getSuggestions(selectedTopicKeywords);
+    } else {
+      setAiSuggestions([]);
+    }
+  }, [selectedTopic, keywordsString]);
```
milljoniaer, line 86: Why cant we do it like this?

```suggestion

  useEffect(() => {
    if (selectedTopic && selectedTopicKeywords.length > 0) {
      getSuggestions(selectedTopicKeywords);
    } else {
      setAiSuggestions([]);
    }
  }, [selectedTopic, selectedTopicKeywords]);
```

frontend/app/custom-components/profile/edit/topics.tsx
```diff
@@ -26,9 +29,74 @@ export function Topics({ profile, setProfile }: TopicsProps) {
...
+
+    setIsLoadingSuggestions(true);
+    try {
+      const result = await client.POST(
+        "/api/v1/search-profiles/keywords/suggestions",
+        {
+          body: keywords,
+          headers: authorizationHeaders,
+        },
+      );
```
milljoniaer, line 58: fyi, there is an open request to change auth to use cookies instead of the authorization Headers, if it gets merged in time, you can/should remove the authorizationHeaders from this request.

milljoniaer (commented):
Nice, thank you guys, almost done from a frontend perspective, can you please have a look at my comments?

And @nikbauer720 what do you think for the backend. Looks good to me

frontend/app/custom-components/profile/edit/ai-suggestion-tag.tsx
neher42: Do you mean the whole tag a button?

nikbauer720 (commented):
Backend looks good!

milljoniaer (approved):
LGTM

PR #179, opened 2025-06-11
Feat/mock breaking news

MagMagnita:
added news carousel
please give feedback on minimalistic design

image

image


MagMagnita:
We also need an issue for this task right?

LeonardoZambranoL:
grafik
- We could reduce some of that whitespace. Also, the card do not quite convey urgency. Use some more agressive colors.

- This has problems with responsiveness. cursor-pointer on the button containing the show more.

grafik
- Expanding them down messes with out layout. This would actually be a good oportunity for us to start using the "article page" we defined at the beginning during figma ideation. We could have a link inside this card, when clicked, we navigate the user to the corresponding article inside the application.

As arefresher, the "article page":
grafik

milljoniaer:
Maybe in order to remove the whitespace left and right on the swiper, you could move the arrows below the cards, more like a pagination

MagMagnita:
Is the blue for breaking news still not alarming enough?

milljoniaer:
nice, like the update!
please dont use blue for the read more links, but standard links

and maybe a bit colored cards could be nice

quick chat with chatgpt, @LeonardoZambranoL @MagMagnita what do you think?
image

MagMagnita:
The light yellow/orange cards look nice for attracting attention, I am not that fan of the orange headers tbh but its the onyl complement color I guess so fine for me :)

milljoniaer:
@LeonardoZambranoL opinions?

LeonardoZambranoL:
That last recommendation looks much more alarming for sure. Maybe go a bit lighter on the background color, not that strong. Headers seem fine.

I would preffer it if all of the "Read more" would be on the same height and all titles had the same number of lines. https://tailwindcss.com/docs/line-clamp should help.

MagMagnita:
The not same height and not same header lines are probably just form jonathan's chatgpt output because mine are all fine

frontend/package.json
```diff
@@ -18,11 +18,12 @@
     "@radix-ui/react-accordion": "1.2.11",
     "@radix-ui/react-alert-dialog": "1.1.14",
     "@radix-ui/react-checkbox": "1.3.2",
+    "@radix-ui/react-collapsible": "1.1.11",
     "@radix-ui/react-dialog": "1.1.14",
     "@radix-ui/react-dropdown-menu": "2.1.15",
     "@radix-ui/react-label": "2.1.7",
     "@radix-ui/react-popover": "1.1.14",
-    "@radix-ui/react-scroll-area": "1.2.9",
+    "@radix-ui/react-scroll-area": "^1.2.9",
```
carlobortolan, line 26: Please pin the dependencies to a specific version, e.g., 
```
"@radix-ui/react-scroll-area": "1.2.9"
```

frontend/app/pages/dashboard/breaking-news.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+type Props = {
+  breakingNews: NewsItem[];
+};
+
+function truncateAtWord(text: string, maxLength: number) {
+  if (text.length <= maxLength) return text;
+  const truncated = text.slice(0, maxLength);
+  return truncated.slice(0, truncated.lastIndexOf(" ")) + "…";
+}
```
carlobortolan, line 26: This could be extracted to a utility function for reuse/testability.

MagMagnita:
image

MagMagnita:
We can merge now, was just a temporary error.

MagMagnita:
Also removed the carousel from the package.json and its tsx file

carlobortolan (approved):
Looks good to me 👍

Minor UI suggestion: You might consider making the entire card clickable to navigate to the article, while keeping the "Show more" text just for expanding the summary.

milljoniaer:
currently the layout is broking because of this.
image
 should be fixed, when you remove the ´grid´ class on the `main` html tag in the Layout component

LeonardoZambranoL (dismissed):

milljoniaer (dismissed):

nikbauer720 (approved):

PR #183, opened 2025-06-11
Feat/search profile deletion and confirmation

MagMagnita:
- Added three dots menue

image

image

- Added delete Dialog (mocked, no endpoint and backend not working right now)

image

- Added cancle Dialog

image


MagMagnita:
Also: We dont have a issue for this task

LeonardoZambranoL:
add some icons

PR #222, opened 2025-06-15
fix: sentry loading

### LINKED ISSUE #218: bug: Stuck on "Oohps" page
To reproduce:

- Login
- From the dashboard, click on "Read more". From one of the breaking news cards.
- You are redirected to "/articlepage" and get a Page Not Found. I tought one would be redirected to "/error/404".
- Click either the logo or the "Back to Home". 
- You are now stuck on a page that only says "Oohps" (wierd spelling btw). No matter where you click, you come back here.

Image

milljoniaer:
- @LeonardoZambranoL I think a reroute to the error 404 route is not really necessary, as this is pretty normal for client-side error boundaries.
- The second error on rerouting to "/" was because the remounting of the app after the error boundary. That triggered reloading sentry. Should be fixed
closes #218


carlobortolan (approved):
Thanks :+1:

PR #248, opened 2025-06-18
244 fix vector store

dancing-shark:
#244 close

### DIFF
--- a/backend/app/api/v1/endpoints/vector_store_controller.py
+++ b/backend/app/api/v1/endpoints/vector_store_controller.py
@@ -1,10 +1,10 @@
 from fastapi import APIRouter, BackgroundTasks, Depends
 
 from app.core.auth import get_authenticated_user
+from app.core.db import get_vector_service
 from app.core.logger import get_logger
 from app.models import User
 from app.services.article_vector_service import ArticleVectorService
-from app.core.db import get_vector_service
 
 router = APIRouter(prefix="/vector-store", tags=["vector-store"])
 
--- a/backend/app/api/v1/endpoints/vector_store_controller.py
+++ b/backend/app/api/v1/endpoints/vector_store_controller.py
@@ -1,8 +1,8 @@
 from fastapi import APIRouter, BackgroundTasks, Depends
 
 from app.core.auth import get_authenticated_user
-from app.core.db import get_vector_service
 from app.core.logger import get_logger
+from app.core.service import get_article_vector_service
 from app.models import User
 from app.services.article_vector_service import ArticleVectorService
 
@@ -16,7 +16,7 @@ async def add_articles_to_vector_store(
     background_tasks: BackgroundTasks,
     page_size: int = 100,
     current_user: User = Depends(get_authenticated_user),
-    vector_service: ArticleVectorService = Depends(get_vector_service),
+    vector_service: ArticleVectorService = Depends(get_article_vector_service),
 ):
     """
     Add a list of articles to the vector store.
--- a/backend/app/api/v1/endpoints/vector_store_controller.py
+++ b/backend/app/api/v1/endpoints/vector_store_controller.py
@@ -4,46 +4,11 @@ from app.core.auth import get_authenticated_user
 from app.core.logger import get_logger
 from app.models import User
 from app.services.article_vector_service import ArticleVectorService
+from app.core.db import get_vector_service
 
 router = APIRouter(prefix="/vector-store", tags=["vector-store"])
 
 logger = get_logger(__name__)
-
-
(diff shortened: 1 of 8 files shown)


backend/app/services/article_vector_service.py
```diff
@@ -46,29 +46,8 @@ def __init__(self):
         )
 
     def create_collection_safe(self, collection_name: str) -> None:
```
ricayanzon, line 48:
```suggestion
    def create_collection(self, collection_name: str) -> None:
```

I don't think we need this here - it was for me when I was working in this file to understand the difference between the two functions (one catches errors while the other can fail). but since you removed one now, it's ok without the `_safe` I think

backend/app/services/article_vector_service.py
```diff
@@ -109,14 +88,15 @@ def add_articles(self, articles: List[Article]) -> None:
...
+                    "id": str(article.id),
+                    "subscription_id": str(article.subscription_id),
+                    "title": article.title,
                 },
             )
             for article in articles
         ]
 
-        uuids = [article.id for article in articles]
+        uuids = [str(article.id) for article in articles]
```
ricayanzon, line 99: you're iterating twice over the `articles` sent and also converting twice all `article.id`s to strings. wouldn't it be better to combine these?

backend/app/services/article_vector_service.py
```diff
@@ -147,32 +127,15 @@ async def index_summarized_articles_to_vector_store(
...
         offset: int = page * page_size
 
-        articles: Sequence[Article] = (
+        articles: List[Article] = (
             await ArticleRepository.list_articles_with_summary(
                 limit=page_size, offset=offset
             )
         )
 
         while articles:
```
ricayanzon, line 136:
```suggestion
        while len(articles) > 0:
```

I know it works in python as was, but it is general cleaner to be specific in your conditions. this would also work if articles is an object, True or other values, which may break other parts that follow after this in the app. for debugging, it's generally easier if the error is catched or breaks something closer to the origin, then a few steps down the line

backend/tests/integration_tests/test_article_vector_service_integration.py
```diff
@@ -0,0 +1,129 @@
...
+        content="",
+        created_at=None,
+        updated_at=None,
+    )
+
+
+@pytest.fixture(scope="module")
+def qdrant_client():
+    client = get_qdrant_connection()
+    # Skip, falls kein Qdrant erreichbar ist
```
ricayanzon, line 31: please, no comments in german. there are a couple of them in this file

backend/app/api/v1/endpoints/vector_store_controller.py
```diff
@@ -10,48 +11,12 @@
...
-        return {
-            "message": "You do not have permission to perform this action."
-        }
-    vector_service.delete_collection(collection_name)
-    return {"message": f"Collection '{collection_name}' deleted successfully."}
-
-
 @router.post("/add-articles")
 async def add_articles_to_vector_store(
     background_tasks: BackgroundTasks,
```
nikbauer720, line 16: Shouldn't the `background_tasks `be defined in the service itself? or is this more for testing? Same for `page_size`?

backend/tests/unit_tests/test_article_vector_service.py
```diff
@@ -0,0 +1,152 @@
...
+        title=title,
+        subscription_id=subscription_id or uuid.uuid4(),
+        content="Original content",
+        created_at=None,
+        updated_at=None,
+    )
+
+
+@pytest.fixture(autouse=True)
+def patch_qdrant_and_store(monkeypatch):
```
nikbauer720, line 26: Wenn wir das generell für Test brauchen (gerade was das matching angeht) sollten wir das fixture in die conftest.py machen? @ricayanzon

nikbauer720 (commented):
Just a few changes :) Thanks for the tests. Than our coverage is better again 👍

nikbauer720 (commented):
How does this PR interfere with the one you are working on right now? Will big parts be reused?

dancing-shark:
> How does this PR interfere with the one you are working on right now? Will big parts be reused?

This PR is for the step embedding articles in the vector store, it is independent from the matching PR, but i use this PR for the similarity search in my matching PR

backend/tests/unit_tests/test_article_vector_service.py
ricayanzon: @nikbauer720 should this be moved to `conftest.py` then? if so, please only resolve the comment once it has been done (I couldn't see a commit implementing this - why did you resolve it?). or did you agree on something else @nikbauer720 ?

PR #260, opened 2025-06-20
delete redundant match_repositoy.py

nikbauer720:
delete the match_repositoy.py and put the function in the right repo


backend/app/repositories/match_repository.py
```diff
@@ -67,3 +69,22 @@ async def update_match_feedback(
...
             await session.refresh(match)
             return match
+
+    @staticmethod
+    async def get_recent_match_count_by_profile_id(
+        profile_id: UUID,
+        since: datetime,
+    ) -> int:
+        async with async_session() as session:
+            since = since.replace(tzinfo=None)
```
ricayanzon, line 79: you should not remove the timezone info, but set it on the missing or both timestamps - this will throw errors again once @specialfish9 has added them to the db timestamps

ricayanzon:
there are also two services: `matching.py` and `matching_service.py`. please remove one of the two, or rename if they are actually doing different stuff.

backend/app/repositories/match_repository.py
nikbauer720: I specified it to be utc.

PR #265, opened 2025-06-21
133 feat implement improved matching logic using qdrant

dancing-shark:


backend/app/core/service.py
```diff
@@ -0,0 +1,8 @@
+from app.services.article_matching_service import ArticleMatchingService
+
+
+def get_article_matching_service() -> ArticleMatchingService:
```
ricayanzon, line 4: are you even using this somewhere?

backend/app/repositories/match_repository.py
```diff
@@ -92,3 +92,14 @@ async def get_recent_match_count_by_profile_id(
             )
             result = await session.execute(stmt)
             return result.scalar_one_or_none() or 0
+
+    @staticmethod
+    async def insert_match(match: Match) -> Match:
+        """
+        Insert a new Match into the database.
+        """
+        async with async_session() as session:
```
nikbauer720, line 101: It is better to have the async_session() at service level. Do you only make this one DB call in the procedure?

milljoniaer (approved):

PR #267, opened 2025-06-22
Feat/extended rights on search profiles

nikbauer720:


backend/app/api/v1/endpoints/user_controller.py
```diff
@@ -47,7 +47,7 @@ async def get_current_user_info(
     return current_user
 
 
-@router.get("/sync", response_model=UserEntity)
+@router.put("/sync", response_model=UserEntity)
```
carlobortolan, line 50: Please update the client side too when merging this (e.g., see #294)

milljoniaer: yes, pls do it in this PR, should be a one line change :)

milljoniaer (commented):
I have one request from a api user perspective:
Can we switch e.g. can_read & readers in the getter of a SP

to can_read being the list of ids and `reader` being the boolean if the user can read?

so that when we are editing, the types are not confusing

backend/app/schemas/search_profile_schemas.py
```diff
@@ -38,8 +40,10 @@ class SearchProfileDetailBase(BaseModel):
     is_public: bool
     organization_emails: List[EmailStr]
     profile_emails: List[EmailStr]
-    editable: bool
-    is_editable: bool
+    can_read: bool
+    readers: List[UUID]
+    can_edit: bool
+    editors: List[UUID]
```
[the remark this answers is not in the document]
nikbauer720: yes, but then you have to evaluate this in the frontend

backend/app/repositories/user_repository.py
```diff
@@ -109,22 +110,21 @@ async def update_user(
...
 
 
-async def get_user_list_from_organization(
-    user: User, session: AsyncSession
-) -> Union[list[User], User]:
-    """
+async def get_user_list_by_org(
+    user: UserEntity, session: AsyncSession
+) -> Union[List[UserEntity], UserEntity]:
+    """Add commentMore actions
```
ricayanzon, line 116: what's this comment? copilot? haha

backend/app/repositories/user_repository.py
```diff
@@ -109,22 +110,21 @@ async def update_user(
...
     Return all users in the same organization, or the user itself if
     no organization.
     Superusers receive all users.
     """
     if user.is_superuser:
-        query = select(User)
+        stmt = select(User).options(selectinload(User.organization))
     elif user.organization_id is None:
         return user
```
ricayanzon, line 124:
```suggestion
    elif user.organization_id is None:
        return [user]
```

you should always return the same type - this way, the frontend needs to manually check: is it only an instance or is it an array/list that is returned. avoid this by simply always returning a list, and if there is only one user: the frontend doesnt care, it iterates over the array/list of users and if there's only one, it shows only one.

backend/app/repositories/user_repository.py
```diff
@@ -109,22 +110,21 @@ async def update_user(
     return _to_user_entity(user)
 
 
-async def get_user_list_from_organization(
-    user: User, session: AsyncSession
-) -> Union[list[User], User]:
-    """
+async def get_user_list_by_org(
+    user: UserEntity, session: AsyncSession
+) -> Union[List[UserEntity], UserEntity]:
```
ricayanzon, line 115:
```suggestion
) -> List[UserEntity]:
```

don't make things more complicated than they have to be: just always return a list (see comment below)

backend/app/services/user_service.py
ricayanzon, line 1: please look at how you imported the repository files in `search_profiles_service.py`: if you import single methods as here, you will:
1) have a lot of import statements from the same files
2) you won't know by the name of the methods, which repository you are using, unless you scroll up and look from where it was imported. i.e. imagine you use two repositories in this service and both have a method `get_by_id`, but you only use one of those. now imagine you are currently working in line 150, so you don't see which one you have imported in the top of the file - you won't really know unless you looked it up previously or you scrolled all the way up just to check - or, if you're lazy or forget that there even is another `get_by_id`, you might think you are using the correct one from one repository, but actually are using the one from the wrong, second repository -> I have seen this in the past. happens to the best. simple solution: import repositories entirely and then do something like `UserRepository.get_user_by_id`.

davidoof3141 (approved):

PR #289, opened 2025-06-24
Feat/translation and entity extraction

AlejandroGuevara05:
Services for both translations and entity extraction with batch api are ready, but I'm having problems with alembic after doing the merge, therefore I open it as a draft


LeonardoZambranoL:
Hi!

Yes, DeepL was experimented with. OpenAI was just much better pricewise. Since we will translate thousands daily, we chose LLMs.

nikbauer720 (commented):
You have to run those commands to format the code: 

black app
isort app
flake8 app

backend/alembic/versions/bdee7e922ed6_translation_columns_in_articles_and_.py
```diff
@@ -0,0 +1,54 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('entities',
+    sa.Column('id', sa.Uuid(), nullable=False),
+    sa.Column('entity_type', sa.Text(), nullable=False),
+    sa.Column('value', sqlmodel.sql.sqltypes.AutoString(length=255), nullable=False),
```
nikbauer720, line 28: if you use nullable = false you need to set a default value.

backend/alembic/versions/bdee7e922ed6_translation_columns_in_articles_and_.py
```diff
@@ -0,0 +1,54 @@
...
+    sa.Column('article_id', sa.Uuid(), nullable=False),
```
nikbauer720, line 29: same here

backend/app/repositories/entity_repository.py
```diff
@@ -0,0 +1,64 @@
+from typing import List, Optional
+from uuid import UUID
+
+from app.core.db import async_session
+from app.models.entity import ArticleEntity
+
```
nikbauer720, line 6: Maybe describe up here what an entity is. The word is not that clear (but I can't make up a better one)

backend/app/services/article_summary_service.py
```diff
@@ -1,105 +1,170 @@
...
+                    article_id, summary
+                )
+                await ArticleEntityRepository.add_entities(
+                    article_id,
+                    persons=persons,
+                    industries=industries,
+                    events=events,
+                    organizations=organizations,
+                )
+                logger.info(
```
nikbauer720, line 119: I don't log when it works, but better when it does not work

backend/alembic/versions/420ae24651f2_add_default_value_to_null_true_columns.py
```diff
@@ -0,0 +1,59 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = '420ae24651f2'
+down_revision: Union[str, None] = 'd72932518738'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
```
nikbauer720, line 23: Please alter the column dircetly in the version it will be created. Otherwise this will still cause the error to happen. First there is no default (error) than the default will be set when it is to late. If you have questions let me now.

nikbauer720 (commented):
Perfect!

PR #294, opened 2025-06-24
Fix sync user API call

carlobortolan:
Fix frontend to use the correct method when calling the `/users/sync` endpoint after #237 changed the endpoint.

This should resolve the current problems when loading the dashboard.

### DIFF
--- a/frontend/app/hooks/use-authorization.tsx
+++ b/frontend/app/hooks/use-authorization.tsx
@@ -61,7 +61,7 @@ export const AuthorizationContextProvider = ({
   useEffect(() => {
     if (isLoaded && isSignedIn) {
       const sync = async () => {
-        const { data: returnedUser } = await client.POST("/api/v1/users/sync");
+        const { data: returnedUser } = await client.GET("/api/v1/users/sync");
         setMediamindUser(returnedUser);
       };
       sync();


nikbauer720:
> @nikbauer720 can you confirm that this is now a `GET` instead of a `POST`?

It is actually quite hard to tell. 

The function takes the clerk user and:
- creates the user if the user does not exists on our db
- or updates the user on our db

at the end it returns the user
What is it now? I would say in most cases it is a Put as on my PR #267

ricayanzon:
it's a POST since it triggeres a process: sync user. but the main question is: what is it in the backend? it fails now because the backend endpoint was changed but it wasn't changed in the frontend, from where it is called

nikbauer720:
> > it's a POST since it triggeres a process: sync user. but the main question is: what is it in the backend? it fails now because the backend endpoint was changed but it wasn't changed in the frontend, from where it is called
> 
> atm it is a GET (changed in #237 from POST -> GET)

Then lets agree on POST? I will change it in the PR #267 accordingly

carlobortolan:
> Then lets agree on POST? I will change it in the PR #267 accordingly

Sounds good to me :+1: 

Do you think that you can get #267 merged soon? If so, I would then close this PR.

MagMagnita (approved):

PR #310, opened 2025-06-25
Fix: regenerated lost alembic files

nikbauer720:
Those two alembic versions are lost but we somehow still need need them: 
`https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/pull/289/files/4c71eb74f3c8a58194aa70fa0174dc71005f973a..e5d8d3b49481787c05c3c7aa3b298cbe12cf0213#diff-3c1053ff48564b4336bdd4eaf98726789526fd43c0b90f4f43968d88efae06b0`

`https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/pull/289/files/4c71eb74f3c8a58194aa70fa0174dc71005f973a..e5d8d3b49481787c05c3c7aa3b298cbe12cf0213#diff-d2c7d8bff741c74014208f7110fb91c889403fd6c5dbc6001bbf49497667fcf9`

### DIFF
--- a/backend/alembic/versions/420ae24651f2_add_default_value_to_null_true_columns.py
+++ b/backend/alembic/versions/420ae24651f2_add_default_value_to_null_true_columns.py
@@ -1,0 +1,42 @@
+"""Add default value to null=true columns
+Revision ID: 420ae24651f2
+Revises: d72932518738
+Create Date: 2025-06-25 19:36:33.800364
+"""
+
+from typing import Sequence, Union
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel.sql.sqltypes
+
+# revision identifiers, used by Alembic.
+revision: str = "420ae24651f2"
+down_revision: Union[str, None] = "d72932518738"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.alter_column(
+        "entities",
+        "id",
+        server_default=sa.text("'00000000-0000-0000-0000-000000000000'"),
+    )
+    op.alter_column(
+        "entities", "entity_type", server_default=sa.text("'unknown'")
+    )
+    op.alter_column("entities", "value", server_default=sa.text("'unknown'"))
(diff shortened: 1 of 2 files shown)


backend/alembic/versions/420ae24651f2_add_default_value_to_null_true_columns.py
```diff
@@ -0,0 +1,42 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.alter_column(
+        "entities",
+        "id",
+        server_default=sa.text("'00000000-0000-0000-0000-000000000000'"),
+    )
```
davidoof3141, line 25: This will also throw errors no? I think there should be a default factory.

backend/alembic/versions/420ae24651f2_add_default_value_to_null_true_columns.py
```diff
@@ -0,0 +1,42 @@
...
+    op.alter_column(
+        "entities", "entity_type", server_default=sa.text("'unknown'")
+    )
+    op.alter_column("entities", "value", server_default=sa.text("'unknown'"))
+    op.alter_column(
+        "entities",
+        "article_id",
+        server_default=sa.text("'00000000-0000-0000-0000-000000000000'"),
+    )
```
davidoof3141, line 34: Is it a good idea to set the foreign key to a dummy value? Won't this throw errors because there are no article_id with this value?

davidoof3141 (changes requested):
Those are the same parameters that we have during the init schema during the creation in this [file](https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/blob/master/backend/alembic/versions/bdee7e922ed6_translation_columns_in_articles_and_.py). Or what is different?

backend/alembic/versions/420ae24651f2_add_default_value_to_null_true_columns.py
davidoof3141: We need to make sure that there is no existing data in "entities" when migrating.

davidoof3141 (approved):

PR #266, opened 2025-06-22
feat/articles explorer endpoint

huaihaochuan:
adapted endpoints of the article overview and the article detail page.
waiting for new schema to merge

PR #314, opened 2025-06-26
fix: adapted schemas

davidoof3141:


backend/app/schemas/articles_schemas.py
```diff
@@ -23,15 +34,25 @@ class ArticleOverviewItem(BaseModel):
...
+            url=getattr(article, "url"),
+            image_url=getattr(article, "image_url", None),
+            authors=getattr(article, "authors", None),
+            published_at=getattr(article, "published_at"),
+            language=getattr(article, "language", None),
+            categories=getattr(article, "categories", None),
+            summary=getattr(article, "summary", None),
+            title_en=getattr(article, "title_en", None),
+            title_de=getattr(article, "title_de", None),
+            content_en=getattr(article, "content_en", None),
```
huaihaochuan, line 49: Nice value handling, we have new schema in [https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/pull/266](url) so you might need to do some adaptions on that

davidoof3141: The link does not work

backend/app/schemas/breaking_news_schemas.py
```diff
@@ -0,0 +1,33 @@
+from datetime import datetime
+from typing import Optional
+
+from pydantic import BaseModel, ConfigDict
+
+from app.models.breaking_news import BreakingNews
+
+
+class BreakingNewsItem(BaseModel):
```
nikbauer720, line 9: should we also make translations for breaking news? But we can do this in another PR

nikbauer720 (approved):

PR #321, opened 2025-06-27
292 matching adapt to multiple matchings per day

### LINKED ISSUE #292: Matching: adapt to multiple matchings per day
Open questions:
- How often should we trigger a matching per day? 
- Can an article appear in multiple matches per day? 
- How do we distinguish between multiple matches per day? ➡ Potential solution: index/counter with highest number representing latest match?

Use Case: If we match an article in the morning which is relevant, should it appear later on again if it is still relevant or only the most recent ones (and than also articles which are less relevant as the most relevant ones have already been matched in the morning if there aren't a load of new ones)

dancing-shark:
close #292


dancing-shark:
image

backend/alembic/versions/469815c7416b_add_matching_runs_tabel.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    op.create_foreign_key(
+        None, "matches", "matching_runs", ["matching_run_id"], ["id"]
+    )
+    op.add_column(
+        "reports", sa.Column("matching_runs_id", sa.Uuid(), nullable=True)
+    )
+    op.create_foreign_key(
+        None, "reports", "matching_runs", ["matching_runs_id"], ["id"]
+    )
```
huaihaochuan, line 68: In the matches table, the name is matching_run_id without the s, would it be better if we keep this consistent?

backend/alembic/versions/469815c7416b_add_matching_runs_tabel.py
```diff
@@ -0,0 +1,83 @@
+"""add matching_runs tabel
```
huaihaochuan, line 1:
```suggestion
"""add matching_runs table
```

backend/app/services/article_matching_service.py
```diff
@@ -238,6 +242,20 @@ async def run(self, batch_size: int = 100) -> None:
             )
         )
 
+        matching_run: MatchingRun
+
+        if len(profiles) > 0:
+            # Create a matching run entry
+            algorithm_version = (
+                f"{self.weights['topic']}_{self.weights['keyword']}"
```
ricayanzon, line 250: oh wow, so algorithm version are the percentages you used for weighting the topic and keyword matching score when combining them? please be more descriptive with your naming - this is not intuitive at all. I thought this were algorithm names like `bm25` or `cosine_similarity`. or `v1` of our own matching algorithm (as e.g. combining different existing algorithms to one). so: 1) it's not very intuitive what this even is, 2) if you go through the database and see entries like `0.7_0.3` - I'm 100% sure that someone will have to look this up in your code to understand what the first value (0.7) stands for (topic) and what the second value (0.3 - keyword) is. so imagine this: you would have to go back to your code, check where the `algorithm_version` was set, to then find out in which order they are placed in the string, just to be sure that 0.7 is actually how much the topic matching is weighted against the keyword matching. super annoying if you ask me. please come up with a better solution for the name(s) and also how you store these values such that you make it easier for anyone that has to work with your code in the future to understand what you meant to do here

nikbauer720:
> > Missed a few matching_runs :)
> 
> I did not. Everywhere in the db there is the s:
> 
> image

you're right but I had a look at the back_populates statements and it seams inconsistent there. Sometimes there is an s and sometimes there is none. Not only for the matching_run(s) but in generell. We should make this consistent.

PR #324, opened 2025-06-28
Feat/migration actions

specialfish9:


.github/workflows/backend-deploy.yml
carlobortolan, line 1: I assume we can do the same for the prod db, right?

```yml
  apply-migrations:
    uses: ./.github/workflows/apply-migrations.yml
    secrets:
      DB_SERVER: ${{ secrets.PROD_DB_SERVER }}
      DB_PORT: ${{ secrets.PROD_DB_PORT }}
      DB_USER: ${{ secrets.PROD_DB_USER }}
      DB_PASSWORD: ${{ secrets.PROD_DB_PASSWORD }}
      DB_NAME: ${{ secrets.PROD_DB_NAME }}
```

(we can also add this in a separate PR, after this is merged, but I think it might make sense to also apply the migrations there)

.github/workflows/apply-migrations.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+          cache: "pip"
+
+      - name: Copy env
+        run: cp .env.example .env
+
+      - name: Run Alembic upgrade
```
carlobortolan, line 36: Don't we need to run `pip install -r requirements.txt` before this?

carlobortolan (changes requested):
Added two small comments; rest looks good to me :+1:

carlobortolan (approved):

.github/workflows/backend-deploy.yml
ricayanzon: let's go automatically with an action. but I've asked chatgpt (since I haven't done this before) and it seems like there's a way without entering DB secrets to GitHub:

Screenshot from 2025-06-30 13-32-35

What do you think?

ricayanzon (approved):

PR #328, opened 2025-06-30
Feat/LLM Login

AlejandroGuevara05:
Current issues:
- Can't go through large htmls (token limit)
- Can go through 2 step login processes (submit username in one page and submit password in another) but doesn't store the second button
- Currently working with a json, not sure of the logic of how we want to integrate with the database
- We need a more robust way of confirming the login. Now the code checks if the username and password were inserted and the submit button was clicked, but there are cases that for some reason you have to click the submit button two times, so we assume we did the login but we did not


AlejandroGuevara05:
I'll solve the mentioned issues and once I have a final version I'll test it with the broken domains, if everything goes fine I'll open the PR properly :)

AlejandroGuevara05:
Implemented the improvements and fixed the login for:
- Handelsblatt - ePaper
- WiWo - ePaper
- Berlin.Table
- HORIZONT+
- Lebensmittelzeitung

Wasn't able to fix:
- FAZ/FASZ - E-Paper
- The Pioneer
It just doesn't select the correct xpaths, I'll see what it can be done, but in the meantime I think this version is fine

PR #339, opened 2025-07-02
Add test controller for chat reply

carlobortolan:
Added test controller used by the lambda function to trigger chat reply.

Current flow:

1. A user sends an email
2. Amazon SES receives the email and stores it in an S3 bucket. (*)
3. The new file upload to S3 triggers an AWS Lambda function
4. The Lambda function retrieves the email content from S3 and sends a POST request to the `/api/v1/chats` endpoint authenticated via an API key (**)  with:
    - the sender’s email address
    - the subject line
    - the message body 
5. Backend receives POST request and queries LLM (currently mocked)
6. Backend sends reply mail

> (* SES does not include the full message body in its notification payload. To access the email content, the message must first be stored in S3 and then read from there...)
> (** This is just to avoid unauthorized users / someone else than our Lambda function to trigger the chats - I've already added the API key to the config.py, notion page, and secrets manager)
>
> Note: Since the Lambda function depends on the endpoint added in this PR, I can only start implementing steps 3&4 after this PR has been merged

Closes #270


backend/app/api/v1/endpoints/chats_controller.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+    # Send a reply email to the user
+    # flake8: noqa: E501
+    email_schedule = EmailSchedule(
+        recipient=user.email,
+        subject=f"[MEDIAMIND] RE: {chat.subject}",
+        content_type="text/HTML",
+        content=f"""<p>Hi {user.first_name},</p>
+        Congratulations on using the slowest, most unsecure and least scalable web chatbot in the world!</p>
```
ricayanzon, line 77: 😄 you dickhead haha

backend/app/api/v1/endpoints/chats_controller.py
ricayanzon, line 1: ok so why exactly is this a test controller? because it only returns a "hello world" message? please implement this as clean as possible - we will build on top of this. so remove the comment, remove the `noqa` comments (it's just splitting a string, that should be less work than writing `flake8: noqa: E501` lol) and make it as clean as possible. the next person will have to remove your mocked response anyway, but comments can be missed and then they just stay... maybe for ever

carlobortolan: Yes sorry for that :sweat_smile: - I've cleaned it up a bit and removed the noqa and test comments

Issue #331, opened 2025-06-30
Chatbot: Store email conversations in DB


ricayanzon:
- [x] create DB table(s) needed (e.g. for chat history)
- [x] store email history in DB
- [x] retrieve email history on new email arrival

PR #351, opened 2025-07-03
Chatbot: Store email conversations in DB

ricayanzon:
Solves #331


carlobortolan (commented):
Looks good to me; some ideas for the new models:
(feel free to disagree / only use those that are necessary)

- We might also want to add some metrics (e.g., token count, total message length, etc.) to limit the length/usage of a conversation (can also be added in a future PR)
- Add a `last_message_at` and `source` field to conversations (in case we want to reuse the chatbot in our web app UI too)
- Add `attachments` (or similar) relationship (if we plan to support files or links to our report model) 
- It might make sense to keep track of the message hierarchy/sequence and store the parent (previous message) and child message (answer) to a certain message

ricayanzon:
> Looks good to me; some ideas for the new models: (feel free to disagree / only use those that are necessary)
> 
> * We might also want to add some metrics (e.g., token count, total message length, etc.) to limit the length/usage of a conversation (can also be added in a future PR)

mm not quite sure what the benefit of having both would be, so token_count and total_message_length. maybe only tokens would be interesting here (since it's what the LLM is costing us). but maybe that's also something that can be calculated on-demand since the entire chat would be loaded anyways --> calculate the tokens count for the total chat then and reply something like: "sorry, token limit exceeded". anyway, I would leave that for a later PR, try to focus now on MVP functionality to speed things up

> * Add a `last_message_at` and `source` field to conversations (in case we want to reuse the chatbot in our web app UI too)

- do we need a `last_message_at` field if we have already the timestamps of all messages and just need to query for the last message with that conversation_id? what would be the benefit of it? maybe I'm not getting it correct
- `source` could be interesting, since we're then not sending an email. agreed. but I would leave this to a follow-up PR (if we get that far in the remaining 3 weeks)

> * Add `attachments` (or similar) relationship (if we plan to support files or links to our report model)

I'd also leave this to a follow-up PR - we don't have much time to implement the chatbot and I'd leave those kind of things for after. now, let's focus on getting the most basic MVP running 

> * It might make sense to keep track of the message hierarchy/sequence and store the parent (previous message) and child message (answer) to a certain message

definitely, but I thought this could be done by sorting the `chat_message`s by `created_at` timestamp - not ideal, but I would try to keep this as simple as possible (performance is not our main goal, for now)

backend/app/models/email_conversation.py
```diff
@@ -0,0 +1,38 @@
...
+            nullable=False,
+            default=datetime.now(timezone.utc),
+        )
+    )
+    updated_at: datetime = Field(
+        sa_column=Column(
+            TIMESTAMP(timezone=True),
+            nullable=False,
+            default=datetime.now(timezone.utc),
+            onupdate=datetime.now(timezone.utc),
```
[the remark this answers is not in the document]
ricayanzon: ah lol I thought that was from my code. sorry too many context switches. I think both are fine tbh. but let me check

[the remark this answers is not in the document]
ricayanzon: the important part is `onupdate` - `func.now()` is using the SQL NOW function under the hood and `datetime.now(timezone.utc)` the python function. I will check if both are valid or only `func.now()`

ricayanzon:
> > ... I would leave that for a later PR, try to focus now on MVP functionality to speed things up
> 
> 👍
> 
> > do we need a last_message_at field
> 
> We'd need it (if we integrate the chatbot also in our webapp) to sort the conversations based on recency without having to query the messages table (currently there's no way to know which conversation was the most recently used one) - but you can leave this (and the other points) for a future PR

we could do this by `updated_at` from `email_conversations` table, right?

> 
> > definitely, but I thought this could be done by sorting the chat_messages by created_at
> 
> Maybe yes, but then we would have to first fetch all messages, sort them by created_at and send them all (which results in very long queries and emails) instead of maybe sending only the last 1-3 messages in the reply email

mm we need to test this. maybe the email clients are smart enough to show you the last messages anyways, so it would not be relevant for email chatbot


> 
> I think having at least the `answer_to` field might be very helpful, but I'll leave this open to you

let's first try to go with a minimalistic approach and add fields when we see that we really need them. maybe you're right - than we can add it later :)

backend/app/models/email_conversation.py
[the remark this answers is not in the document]
ricayanzon: it seems like for PostgreSQL, only `func.now()` works - good catch! changing now

ricayanzon:
> > we could do this by updated_at from email_conversations table, right?
> 
> Afaik the `email_conversations` table won't be updated, when there's a new message (instead, only the `chat_messages` table will be updated) - `email_conversations` only has a FK to `chat_messages` and that doesn't change onupdate

hm good point. we might need to update it "manually" then - whenever a new `chat_message` is added for that `email_conversation`, we need to update the `updated_at` manually.

docs/EMAIL_CHATBOT.md
carlobortolan, line 1: Just FYI: I'll add the SES config & Lambda section (as well as the actual code for our Lambda function) in a separate PR later today

backend/app/services/chatbot_service.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        chatbot_response = f"""<p>Hi {user_first_name},</p>
+            <div style="background-color: #f5f5f5; padding: 10px;
+                margin: 10px 0;">
+                <strong>Conversation History:</strong><br>
+                <pre>{conversation_context}</pre>
+            </div>
+            <p><strong>Your latest message:</strong> {chat.body}</p>
+            <p><strong>Subject:</strong> {chat.subject}</p>
+            <p>This is a mock response.</p>
```
carlobortolan, line 65: I preferred the previous chatbot reply tbh, but this is also ok.
image

backend/app/schemas/chatbot_schemas.py
```diff
@@ -0,0 +1,9 @@
+from pydantic import BaseModel
+
+
+class ChatRequest(BaseModel):
+    sender: str
+    subject: str
+    body: str
+    s3_key: str
+    bucket: str
```
carlobortolan, line 9: Just FYI: I initially added these to the request, in case we want to load the user's chat email-file from S3. However, we probably don't need that as long as we pass the entire content of the mail in the `body` field.

backend/app/services/chatbot_service.py
carlobortolan: ... also, iirc, we talked a couple of weeks ago about Jinja templates; do we already have them? If so, we should put this code there (otherwise just leave it as is, and create a TODO/issue)

carlobortolan (approved):
Looks good to me :+1: 

One thing that's missing is rate limiting (e.g., either limit the length of conversations or the number of new chats per hour; but we can also have that as a separate issue)

ricayanzon:
I'd do it in a separate branch, I'll create an issue. thanks!

docs/EMAIL_CHATBOT.md
carlobortolan: @ricayanzon: Just added them, see: https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/blob/docs/ses-lambda/docs/SES.md

ricayanzon: very nice!

backend/app/schemas/chatbot_schemas.py
ricayanzon: can you then please remove it in your own branch?

carlobortolan (approved):

PR #371, opened 2025-07-04
Fixes in report sending

AlejandroGuevara05:
Fix
* Reports are generated in both english and german
* Reports aren't sent twice to the same user

Now the workflow is:
1. Generates the report for each search profile in english and german
2. Send the report to the emails in the search profile, using for each the language of the user associated to that email


backend/app/services/email_service.py
```diff
@@ -185,20 +186,43 @@ def _build_email_content(
...
+                    user = await UserService.get_by_email(email)
+                    report_in_user_lang = (
+                        EmailService._get_report_in_user_language(
+                            reports, user
+                        )
                     )
+
+                    report = report_in_user_lang["report"]
+                    presigned_url = report_in_user_lang["presigned_url"]
+                    dashboard_url = report_in_user_lang["dashboard_url"]
```
nikbauer720, line 222: Doesn't this work this way?
```suggestion
                    dashboard_url = report_in_user_lang.dashboard_url
```

backend/app/services/report_service.py
```diff
@@ -103,10 +103,10 @@ async def get_report_by_id(report_id: UUID) -> Optional[Report]:
     @staticmethod
     async def run(
         timeslot: str,
-        language: str,
+        languages: List[str],
```
nikbauer720, line 106: Maybe we should implement a enum class for languages so we can use it with auto completion
 
```suggestion
        class Language(str, Enum):
          EN = "en"
          DE = "de"
```
and than use it like this: 

```suggestion
        language == Language.DE
```

This isn't only for your code but to the whole project. Just came to me now. Maybe we should do a fix as an own PR

nikbauer720:
> Added a class for the languages and using it now in the pipeline. This is what you had in mind?

Yes! Nice :)

nikbauer720 (approved):

PR #379, opened 2025-07-05
Squash alembic versions

nikbauer720:
Waiting until there are no further migrations in a PR


davidoof3141:
Which branches are we waiting for?

ricayanzon:
> Which branches are we waiting for?

since this is not urgent, we could also wait until we're done with the chatbot tables. that shouldn't be too long.

PR #381, opened 2025-07-05
Fix: Backend flake8 

nikbauer720:
exclude alembic from flake8 and formatted tests

@ricayanzon is excluding alembic okay? we will squash them anyway so maybe from then on as there are a lot of flake8 issues...


davidoof3141:
I thought in the ci it is only gonna take the /app folder into consideration no?

ricayanzon:
> I thought in the ci it is only gonna take the /app folder into consideration no?

I think it should format/lint all Python files we create. only automatically generated files and maybe some config stuff can be excluded.

ricayanzon (commented):
I think we should also move the formatter up? (black)

nikbauer720:
> I think we should also move the formatter up? (black)

We can simply type `black tests` and it formats the tests.

ricayanzon:
> > I think we should also move the formatter up? (black)
> 
> 
> 
> We can simply type `black tests` and it formats the tests. 

yes but then it doesn't format the tests automatically for those who run it on-save (and never manually) 🫣

nikbauer720:
> > > > > I think we should also move the formatter up? (black)
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > We can simply type `black tests` and it formats the tests.
> > 
> > 
> > > 
> > 
> > 
> > > yes but then it doesn't format the tests automatically for those who run it on-save (and never manually) 🫣
> > 
> > 
> > If you have it on save, it does format every file you are working on, doesn't matter if it is a test or another file.
> 
> yes but it usually looks for a config file in root to let the formatter know how to format the files (on-save). does running 'black test' really also format the test files as specified in the config inside backend?

Do we have a config for black rn? I didn't find one

nikbauer720:
In the current pyproject.toml we have this and it is not specified for app only so it should work for every package inside of backend: 

```
[tool.black]
line-length = 79
target-version = ["py311", "py312", "py313"]
include = '\.pyi?$'
exclude = '''
/(
    \.venv
  | __pycache__
  | build
  | dist
  | venv
  | env
)/
'''
[tool.isort]
profile = "black"
known_first_party = ["app"]
line_length = 79
skip = [
    ".venv",
    "__pycache__",
    "build",
    "dist",
    "venv",
    "env",
]
[tool.pytest.ini_options]
pythonpath = ["."]
asyncio_mode = "auto"
```

PR #394, opened 2025-07-07
Feat/299 pdf improve design

NicolasArteaga:
**Big PDF Reportlab PR**

(**Do not merge yet**, the articles pdf has still feedback from Hao that should be implemented and the Appendix its showing the translated and not yet the original, article)

But it would be great to already get feedback on:
- Refactored the Parts of the PDF creation to be in different files
- The parser for the Markdown to be readable by Trafilatura, under services/markdown_utils.py
- Created a crawl_stats_pdf_service that creates a pdf for Monitoring of the crawlers

Thank you!


backend/app/services/locales/de/LC_MESSAGES/translations.po
```diff
@@ -23,7 +23,7 @@ msgid "Full Article"
 msgstr "Vollständiger Artikel"
 
 msgid "Read full article"
-msgstr "Vollständigen Artikel lesen"
+msgstr "Vollständiger Artikel lesen"
```
nikbauer720, line 26: Vollständigen is actually right in this case

backend/app/services/locales/de/LC_MESSAGES/translations.po
```diff
@@ -8,7 +8,7 @@ msgid "Page"
 msgstr "Seite"
 
 msgid "Daily News Report"
-msgstr "Täglicher Nachrichtenbericht"
+msgstr "Aktueller Pressespiegel"
```
nikbauer720, line 11: Switched away from daily as it is 3x a day?

backend/app/services/locales/de/LC_MESSAGES/translations.po
LeonardoZambranoL: Beautiful language.

davidoof3141 (approved):

PR #395, opened 2025-07-07
Feat/setup proxy servers

davidoof3141:
Feature:
- added connection through authentication proxy servers
- added RSS Crawlers

### DIFF
--- a/backend/app/services/seeding_service/seeding_credentials.py
+++ b/backend/app/services/seeding_service/seeding_credentials.py
@@ -4,7 +4,6 @@ from typing import Any, Dict
 
 from sqlalchemy.future import select
 
-from app.core.config import configs
 from app.core.db import async_session
 from app.core.logger import get_logger
 from app.models.subscription import Subscription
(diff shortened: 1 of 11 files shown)


backend/app/services/seeding_service/seeding_credentials.py
```diff
@@ -77,7 +76,7 @@ async def seed_credentials_from_env() -> None:
     Seed subscription credentials.
     This is used in production deployments.
     """
-    subscription_accounts = configs.SUBSCRIPTION_ACCOUNTS
+    subscription_accounts = None
     if not subscription_accounts:
         logger.info("No SUBSCRIPTION_ACCOUNTS environment variable found")
         return
```
nikbauer720, line 82: Why you setting it None and then checking if it is None?

davidoof3141: I set it none bcs we dont have that env var anymore. I just commented it out now until we clarify how we deal with the seeding

nikbauer720 (dismissed):

nikbauer720 (approved):

PR #404, opened 2025-07-07
fix: get matched articles for profile using last matching run

nikbauer720:


davidoof3141:
@nikbauer720 @dancing-shark 
Is there a reason why this is still in draft?
Just for my understanding: 
We match based on the date but the issue is when multiple matches happened during the timespan right?
Is an article match dependant of another match or how would the matching score change during different runs?

davidoof3141 (approved):

PR #409, opened 2025-07-07
fix: update reports to be sent using chatbot email

carlobortolan:
image

### DIFF
--- a/backend/app/services/email_service.py
+++ b/backend/app/services/email_service.py
@@ -59,7 +59,7 @@ class EmailService:
         for email in emails:
             try:
                 email.attempts += 1
-                EmailService.__send_email(email)
+                EmailService.send_ses_email(email)
                 email.state = EmailState.SENT
                 await EmailRepository.update_email(email)
             except Exception as e:
@@ -128,7 +128,7 @@ class EmailService:
             raise
 
     @staticmethod
-    def _build_email_content(
+    def build_email_content(
         s3_link: str,
         dashboard_link: str,
         profile_name: str,
@@ -238,7 +238,7 @@ class EmailService:
                         recipient=email,
                         subject=subject,
                         content_type="text/HTML",
-                        content=EmailService._build_email_content(
+                        content=EmailService.build_email_content(
                             presigned_url,
                             dashboard_url,
                             search_profile.name,
--- a/backend/app/api/v1/endpoints/email_controller.py
+++ b/backend/app/api/v1/endpoints/email_controller.py
@@ -32,8 +32,11 @@ async def trigger_email_sending(recipient_email: str):
         recipient=recipient_email,
         subject="[MEDIAMIND] Your daily report",
         content_type="text/HTML",
-        content=build_email_content(
-            "<test_s3_link>", "<test_dashboard_link>", "<Test Search Profile>"
+        content=EmailService.build_email_content(
+            "<test_s3_link>",
+            "<test_dashboard_link>",
+            "<Test Search Profile>",
+            "<Test Last Name>",
         ),
     )
 


backend/app/services/email_service.py
```diff
@@ -59,7 +59,7 @@ async def send_scheduled_emails():
         for email in emails:
             try:
                 email.attempts += 1
-                EmailService.__send_email(email)
```
ricayanzon, line 62: is __send_email still used?

carlobortolan: It should be used whenever we want to send a non-chatbot email; e.g., for notifications, sending the scraper stats to the admins, ... (depends on what features we plan to have / already have)

carlobortolan: Also I just noticed that there might be an issue with sending the reports using the chatbot email:

carlobortolan: @ricayanzon 
image

ricayanzon: Let's just do it.

ricayanzon (approved):

backend/app/services/email_service.py
carlobortolan: Ok - but which website & contacts should I put into the verification request?

https://csee.tech?

carlobortolan: I'll wait to merge this until our verification request has been approved - otherwise the report emails would stop working

ricayanzon: what's the status on that? can we merge this?

carlobortolan: Nope not yet - AWS still hasn't accepted our request to enable production access to SES

apps/api/app/services/email_service.py
```diff
@@ -29,6 +28,15 @@
...
+@dataclass
+class EmailServer:
+    hostname: str
+    port: int
+    use_tls: bool
+    user: str
+    password: str
+
+
 class EmailSchedule:
```
specialfish9, line 40: Can you make also this class a dataclass please?

apps/api/app/services/email_service.py
```diff
@@ -159,7 +170,7 @@ def _render_email_template(template_name: str, context: dict) -> str:
             raise
 
     @staticmethod
-    def _build_email_content(
+    def build_email_content(
```
specialfish9, line 173: Why this became public? Is it accessed somewhere else? 🤔

apps/api/app/services/email_service.py
ricayanzon: while doing that, I realized there are tests that have also an attachment field - which the current class doesn't have. do you know why?
Screenshot from 2025-07-16 08-58-23

apps/api/app/services/email_service.py
ricayanzon: @carlobortolan ?

carlobortolan: It used to be public because (if I'm not mistaken) someone used it in some backend service for testing. But I just updated this branch by pulling the latest version of master, and now it's private again, so someone else has already fixed this.

carlobortolan: It used to be public because (if I'm not mistaken) it was called in the email_service.py; but being private, it threw an error when doing so.

Since the email service does not exist anymore (it still existed when I opened this PR), I've now set it to private.

apps/api/app/services/email_service.py
nikbauer720: Can you specify? Now we have attachments again with they are Optional. Based on a boolean we send the PDF as an attachment or as a link

PR #421, opened 2025-07-09
Improvements to llm login approach

AlejandroGuevara05:
Small but critical improvements!
* Confirm login with 200 Response and not with urls comparison
* Cleaning SVG data from html (represents around 100 thousand tokens)
* Handle exceptions in concurrent calls
* Reinsert password if username failed (was causing problems)


davidoof3141:
Btw how is it triggered in the pipeline? Or what is the plan? Theoretically we log it.

AlejandroGuevara05:
I think we should only call it if the hardcoded_login() call fails

nikbauer720 (approved):

PR #424, opened 2025-07-09
Fix/pipeline fixes

davidoof3141:
Multiple bug fixes:
- matching run id validation error
- When email in organization email but not as user -> user is none -> multiple nullpointers
- organization emails contained null values and empty arrays -> default db model to empty array -> manually changed db values in the production db
- hidden user for jonathan in user db with duplicate email caused scalar error
- one function got too many arguments


davidoof3141:
Please do a local test run in case I missed something

backend/app/services/email_service.py
```diff
@@ -191,14 +191,16 @@ def _build_email_content(
 
     @staticmethod
     def _get_report_in_user_language(reports, user):
-        english_report = None
+        if user:
```
ricayanzon, line 194: be more explicit in your conditions: which values can user take? what would be the condition here - any truthy value or a specific truthy value?

backend/app/services/email_service.py
```diff
@@ -191,14 +191,16 @@ def _build_email_content(
...
         for report in reports:
-            if report["report"].language == user.language:
-                return report
             if report["report"].language == "en":
-                english_report = report
-        # If there is no report in the user language
-        return english_report
+                return report
+        # If no English report is found, return first available report
+        return reports[0] if reports else None
```
ricayanzon, line 203:
```suggestion
        return reports[0] if reports is not None and len(reports) > 0 else None
```
be more specific

backend/app/services/email_service.py
```diff
@@ -219,13 +221,12 @@ async def run(reports_infos: List[dict]):
...
                         )
                     )
-
                     report = report_in_user_lang["report"]
                     presigned_url = report_in_user_lang["presigned_url"]
                     dashboard_url = report_in_user_lang["dashboard_url"]
 
                     translator = ArticleTranslationService.get_translator(
-                        user.language
+                        user.language if user else "en"
```
ricayanzon, line 229: please be more explicit in your condition

backend/app/services/email_service.py
```diff
@@ -242,8 +243,8 @@ async def run(reports_infos: List[dict]):
                             presigned_url,
                             dashboard_url,
                             search_profile.name,
-                            user.last_name,
-                            user.language,
+                            user.last_name if user else None,
+                            user.language if user else "en",
```
ricayanzon, line 247: see above

nikbauer720 (approved):

nikbauer720 (approved):

Issue #253, opened 2025-06-19
Scheduler: Connect to reports & emails pipeline


ricayanzon:
(@ricayanzon)
Scheduler should be implemented in #99. Use to create reports and send out emails automatically.

- [ ] replace `backend/app/api/v1/endpoints/email_controller.py` by a scheduling service and remove controller and related files completely from main backend (see resolved comments in #239 regarding that file)
- [ ] use the `time_slot`s defined in #239 for the reports table and adapt accordingly (i.e. `morning`, `afternoon`, `evening`)

(@carlobortolan)
The pipeline could look a bit like this:

1. Schedule report creation 3 times a day (i.e. `morning`, `afternoon`, `evening`)
2. When a create report event is triggered -> create for each search profile a report: 
    ```py
    get_or_create_report(search_profile_id, timeslot) -> Optional[Report]
    ```
4. Schedule emails to be sent to all users who have subscribed to that report's search profile (incl. `recipient` and `report_id`)
5. When a send email event is triggered given a `recipient` and `report_id` -> send email to recipient with link:    
   ```py
   presigned_url = S3Service.generate_presigned_url(key=report.s3_key, expires_in=604800)
    ```

PR #425, opened 2025-07-09
Feat/scheduler

### LINKED ISSUE #99: Setup scheduler on Redis with rq
- Use [rq](https://github.com/rq/rq?tab=readme-ov-file) for queueing jobs and processing them in the background with workers
- Use [rq-dashboard-fast](https://github.com/Hannes221/rq-dashboard-fast) for monitoring rq queues, jobs, and workers in real-time

specialfish9:
Adds scheduler, as discussed in meta weekly

closes #99  #253


scheduler/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    jq \
+    && rm -rf /var/lib/apt/lists/*
+
+# Copy installed Python packages from builder
+COPY --from=builder /install /usr/local
+
+# Create logs directory and set proper permissions
+RUN mkdir -p /app/logs && \
+    chown -R appuser:appuser /app && \
+    chmod -R 755 /app/logs
```
carlobortolan, line 44: Do we need this? Since we use AWS CloudWatch, the console logs should automatically be saved there.
Also, this might crash the container if the log amount over time and occupy more than the allowed 20GB of disk for our ECS service.

scheduler/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    
+# Set permissions and switch to non-root user
+RUN chown -R appuser:appuser /app
+USER appuser
+
+# Expose port if needed (e.g. for web app)
+EXPOSE 8000
+
+ENTRYPOINT ["/entrypoint.sh"]
+CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "8000"]
```
carlobortolan, line 61: Do we want to use a different port here (e.g., `8001`) in case Eutop decides to self-host both the backend and scheduler on the same machine?

scheduler/job.py
```diff
@@ -0,0 +1,32 @@
+import requests
+from config import Config
+from service import SchedulerService
+from uuid import UUID
+import logging
+
+logger = logging.getLogger(__name__)
+
+def schedule_jobs(service: SchedulerService, cfg: Config) -> None:
+    print("Scheduling jobs...", cfg.EMAIL_JOB_INTERVAL)
```
carlobortolan, line 10: Shouldn't we use `logger.info(...)` instead?

scheduler/.env.example
```diff
@@ -0,0 +1,14 @@
+REDIS_HOST=localhost
+REDIS_PORT=6379
+REDIS_DB=0
+REDIS_PASSWORD=secret
```
carlobortolan, line 4: Could we simplify this to be just: `REDIS_URL=rediss://username:password@host:port/db`?

This would make this consistent with the backend's .env and Redis config.

scheduler/config.py
```diff
@@ -0,0 +1,21 @@
+from pydantic_settings import BaseSettings
+from typing import Optional
+
+class Config(BaseSettings):
+    REDIS_HOST: str
+    REDIS_PORT: int
+    REDIS_DB: int
+    REDIS_PASSWORD: Optional[str] = None
```
carlobortolan, line 8: Let's use a single redis URL to keep cleaner (same as backend - see comment above)  
```py
REDIS_URL: str
```

scheduler/job.py
```diff
@@ -0,0 +1,32 @@
...
+
+    service.schedule_periodic(
+        id = UUID("f47ac10b-58cc-4372-a567-0e02b2c3d479"),
+        every_seconds= cfg.EMAIL_JOB_INTERVAL,
+        func = job_request,
+        args =[f"{cfg.API_BASE_URL}/v1/jobs/email"],
+    )
+
+    service.schedule_periodic(
+        id = UUID("4b986edc-360e-4690-8633-ff2f11cdfe3a"),
```
carlobortolan, line 20: Just out of curiosity - what are these UUIDs for?

scheduler/main.py
```diff
@@ -0,0 +1,40 @@
...
+   scheduler.run()  
+
+def main():
+    logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+    cfg = Config()
+
+    redis = Redis(host=cfg.REDIS_HOST, port=cfg.REDIS_PORT, db=cfg.REDIS_DB, password=cfg.REDIS_PASSWORD)
+    logger.info(f"Connected to Redis at {cfg.REDIS_HOST}:{cfg.REDIS_PORT}")
+
+    service = SchedulerService(redis)
```
carlobortolan, line 29: See comment above:

```py
import redis
client = redis.from_url(configs.REDIS_URL)
client.ping()
logger.info(f"Connected to Redis at {cfg.REDIS_HOST}:{cfg.REDIS_PORT}")
service = SchedulerService(client)
```

carlobortolan (changes requested):
I haven't been able to test it locally yet - will do that tomorrow; in the meantime, I've added some code and config-related comments.

But the rest looks good to me :+1:

specialfish9:
> You moved the entire scheduler out of the backend folder - but it is part of the backend (isn't it?).

Yes of course

> why did you place it outside?

Because it feels weird to me to have a project inside the project. Even if it is a backend service, I would keep it in a separate folder.

>  Also, please create a short reference to the scheduler in the backend README and link it's README also there. also, maybe put the scheduler in the `docker-compose.dev.yml`, for completion?

Sure! I forgot :p

specialfish9:
> > Technically it is a separate microservice (which also runs on a separate instance) and independet of the backend (=not part of the `backend/` microservice) - so I think it makes sense to leave it in a separate directory
> 
> in a microservice architecture, you'd still put all microservices inside the backend folder (they all belong to the backend)
> 
> > Because it feels weird to me to have a project inside the project. Even if it is a backend service, I would keep it in a separate folder.
> 
> sure. but I think it's better to move the backend service (and only that; not the config and readme files, e.g.) into it's own folder (call it "main" for main service) inside the backend folder, and move the scheduler service also inside the backend folder.


The second option sounds good 👍🏼  . However, I would not call the folder "main". I think "api" or "rest-api" are more commonly used

carlobortolan:
> in a microservice architecture, you'd still put all microservices inside the backend folder (they all belong to the backend)

I would disagree with that. Usually, each microservice has its own repository - since we're using a mono-repo, all microservices go on the top-level directory.

If all microservices had to go into one folder, our frontend would also have to go into the backend directory, since it runs a node.js server in the background.

Some examples I found: , https://github.com/directus/directus, https://github.com/tum-Dev/gocast (maybe also https://github.com/twitter/the-algorithm)

Alternatively, some large mono-repo projects put all microservices into a `services` or `apps` folder (e.g., https://github.com/supabase/supabase).

backend/app/api/v1/endpoints/jobs_controller.py
```diff
@@ -1,31 +1,48 @@
...
-@router.get("/email")
-async def email_job():
-    await EmailService.send_scheduled_emails()
-
-@router.get("/pipeline")
-async def trigger_pipeline():
-    datetime_start: datetime = datetime.combine(date.today(), datetime.min.time())
-    datetime_end: datetime = datetime.now()
+@router.post("/email")
+async def email_job() -> JobResponse:
```
ricayanzon, line 16: thanks, looks much better now <3

carlobortolan:
tl;dr: we'll move the `backend`, `scheduler` and `frontend` into a `apps/` directory

specialfish9:
After an intense discussion, @ricayanzon , @carlobortolan  and me agreed on an `apps` folder with the following subfolders
- `frontend`
- `api` (previously `backend`)
- `scheduler`

carlobortolan (commented):
I looked at the changes and they seem ok - I updated our Vercel config (on vercel.com) according to our new project structure - other than lets hope this PR doesn't break anything else.

(Consider this PR 'approved' from my side - I commented only to prevent someone accidentaly merging it before @ricayanzon confirms that this doesn't break anything (e.g., the api type ci/cd)

carlobortolan (approved):
:rocket:

PR #427, opened 2025-07-10
feat: added email notification for relevant breaking news

davidoof3141:


davidoof3141:
image

davidoof3141:
> Is it intended that breaking news mails are sent 7.5h after the event happened (or was this just for testing)? 😅
> 
> 

Yes and no. ^^
We (should) fetch all breaking news of the day from newsapi every hour or even less wit the scheduler. 
It is only possible to filter by date and not by time so with each requests we will also get stuff that we already have. And the function will only send emails for breaking news that are new within that interval e.g. 1hour. The test email was a random sample. :)

LeonardoZambranoL:
Only people with an account get the emails. Correct?

backend/app/services/web_harvester/breaking_news_crawler.py
```diff
@@ -219,6 +224,31 @@ def fetch_breaking_news_newsapi():
...
+                    for user in users:
+                        email_schedule = EmailSchedule(
+                            user.email,
+                            f"Breaking News Alert: {article.title}",
+                            "text/plain",
+                            email_content,
+                        )
+                        # Schedule the email
+                        await EmailService.schedule_email(email_schedule)
+                        await EmailService.send_scheduled_emails()
```
carlobortolan, line 249: Currently, this sends an email, waits for it to be sent, and then repeats this for the next user. Since all breaking news emails look the same, I think it would make sense to:

1. Collect all user emails into a list: `emails = [user.email for user in users]`
2. Set them as bcc recipients, rather than sending individual emails in a loop.

That way, we only send out one email in total instead of one for every user.

nikbauer720: But is the GDPR conform as there could be users from other organizations? I think the idea is good but maybe make it one email per organization

davidoof3141: @carlobortolan I actually dont know if the email will stay the same for everyone. I thought we could keep it like that and add a follow up ticket once we get more feedback from hao?

carlobortolan: With the current implementation it is the same email for everyone so I would suggest implementing it this way.

Also we shouldn't be using `schedule_email(email_schedule)` and `send_scheduled_emails()` as this saves the email in our db, and is intended for scheduling the reports (afaik this method should only be called by the scheduler).

Instead we probably want to send the email directly (e.g., with `send_email()`)

PR #436, opened 2025-07-10
Fix filterbar reports disappears when no matches

neher42:


frontend/app/pages/reports/reports.tsx
```diff
@@ -54,7 +54,7 @@ export function ReportsPage() {
     if (!languageFilter) {
       return reports.reports;
     }
-    const langToFilter = languageFilter === "en" ? "us" : "de";
+    const langToFilter = languageFilter === "en" ? "en" : "de";
```
milljoniaer, line 57: what is the latest thing here? Can we be sure that we need `en`?

milljoniaer (approved):

PR #438, opened 2025-07-10
fix: get all subscriptions for organizations

nikbauer720:
Now in get all there are already the subscriptions included as in the search profiles

Closes #439


backend/app/repositories/subscription_repository.py
```diff
@@ -50,6 +50,32 @@ async def get_all_subscriptions_with_search_profile(
...
+            .label("is_subscribed"),
+        )
+
+        result = await session.execute(stmt)
+
+        return [
+            SubscriptionSummary(
+                id=row.id, name=row.name, is_subscribed=row.is_subscribed
+            )
+            for row in result
```
milljoniaer, line 76: This still gets us all subs, but sometimes "is_subscribed" is false, correct?

milljoniaer (approved):

PR #457, opened 2025-07-11
Add breaking news flag to user

### LINKED ISSUE #456: Add Breaking News Flag to users.
This should make it possible for users to decide if they want to receive breaking news emails or not.

nikbauer720:
@davidoof3141 please use this flag to only send it to users with breaking_news = True

closes #456


apps/api/alembic/versions/36d16a1fbaec_add_breaking_news_to_users.py
```diff
@@ -0,0 +1,36 @@
...
+down_revision: Union[str, None] = "7af6cf813735"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "users", sa.Column("breaking_news", sa.Boolean(), nullable=False)
```
davidoof3141, line 27: You need to add a default value

nikbauer720:
> @nikbauer720 How am I supposed to use this? Should I wait for this merge and then use it in my other branch?

Yes, I think it is quite easy as you have have to change the user repository function to add this condition (but then also change the naming of the function)

davidoof3141 (approved):

PR #462, opened 2025-07-11
Add reports skeleton

neher42:
image

### DIFF
--- a/apps/frontend/app/custom-components/breadcrumb-skeleton.tsx
+++ b/apps/frontend/app/custom-components/breadcrumb-skeleton.tsx
@@ -1,0 +1,5 @@
+import React from "react";
+
+export function BreadcrumbSkeleton() {
+  return <div className="h-4 bg-gray-200 rounded animate-pulse w-24"></div>;
+}
--- a/apps/frontend/app/custom-components/reports/report-card-skeleton.tsx
+++ b/apps/frontend/app/custom-components/reports/report-card-skeleton.tsx
@@ -1,0 +1,25 @@
+import { Skeleton } from "~/components/ui/skeleton";
+
+export function ReportCardSkeleton() {
+  return (
+    <div className="shadow-lg rounded-2xl p-4 h-40 w-50 space-y-1 flex flex-col justify-between">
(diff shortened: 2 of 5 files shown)


frontend/app/pages/reports/reports.tsx
```diff
@@ -118,7 +119,7 @@ export function ReportsPage() {
     return pages;
   };
 
-  if (profileLoading || isLoading) {
+  if (profileLoading) {
     return <></>;
   }
```
milljoniaer, line 124: can we please also show the loading cards, while the profile is loading?

frontend/app/pages/reports/reports.tsx
```diff
@@ -156,12 +157,18 @@ export function ReportsPage() {
       <Text hierachy={2}>{t("reports.header")}</Text>
 
       {isLoading ? (
-        <div className="flex items-center justify-center min-h-[400px]">
-          <div className="text-center">
-            <div className="animate-spin rounded-full h-8 w-8 border-b-2 border-gray-900 mx-auto mb-4"></div>
-            <p>{t("reports.loading")}</p>
```
milljoniaer, line 162: Is this translation string still needed? If not please remove

milljoniaer (commented):
Nice, very cool feature, pls have a look at the comment

frontend/app/pages/reports/reports.tsx
neher42: I need the profile for the breadcrumb. So when the profile doesn't exist yet i won't have the breadcrumb. 
So i could show the skeleton but then the breadcrumb would just pop in later.
Not sure what the best way to handle this would be, do you have an idea?

milljoniaer: You can also add a skeleton for the breadcrumb, no?

neher42: Ah ok, i'll do that then.

milljoniaer (changes requested):
Nvmd pls add loading & skeleton for breadcrumb as well

milljoniaer (approved):

PR #463, opened 2025-07-11
Add markdown support article page

neher42:
image

### DIFF
--- a/frontend/app/app.css
+++ b/frontend/app/app.css
@@ -14,6 +14,10 @@ body {
   @apply bg-white dark:bg-gray-950;
 }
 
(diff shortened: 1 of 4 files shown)


frontend/app/app.css
```diff
@@ -14,6 +14,10 @@ body {
   @apply bg-white dark:bg-gray-950;
 }
 
+.markdown > * {
+  all: revert;
+}
```
milljoniaer, line 19: what is that rule needed for?

milljoniaer (approved):
Nice looks good, just one question @neher42

PR #464, opened 2025-07-11
Include citations in endpoint

AlejandroGuevara05:
Included the article's citations in the endpoint /api/v1/search-profiles/{search_profile_id}/article/{match_id}, so the frontend can show them


AlejandroGuevara05:
All the entities are being sent in another pr

PR #486, opened 2025-07-13
feat: scheduler infra

specialfish9:
Leftover of #425

### DIFF
--- a/infra/main.tf
+++ b/infra/main.tf
@@ -102,6 +102,19 @@ module "ecs_dev" {
   alb_security_group_id = module.alb.alb_security_group_id
 }
 
+module "scheduler" {
+  source                = "./modules/scheduler"
+  service_name          = "mediamind-service-scheduler-dev"
+  cluster_name          = local.cluster_name
+  container_image       = module.ecr.repository_url
+  redis_endpoint        = module.redis.endpoint
+  subnet_ids            = data.aws_subnets.selected.ids
+  vpc_id                = data.aws_vpc.selected.id
+  secrets_arn           = local.dev_secrets_arn
+  s3_backend_bucket     = module.s3.bucket
+  region                = "eu-central-1"
+}
+
 module "qdrant" {
   source         = "./modules/qdrant"
   service_name   = "mediamind"
--- a/infra/modules/scheduler/main.tf
+++ b/infra/modules/scheduler/main.tf
@@ -1,0 +1,115 @@
+variable "cluster_name" { type = string }
+variable "service_name" { type = string }
+variable "container_image" { type = string }
+variable "redis_endpoint" { type = string }
+variable "subnet_ids" { type = list(string) }
+variable "vpc_id" { type = string }
+variable "secrets_arn" { type = string }
+variable "s3_backend_bucket" { type = string }
+variable "region" { type = string }
+variable "log_group_name" {
+  type    = string
+  default = null
+}
+
+locals {
+  effective_log_group_name = var.log_group_name != null ? var.log_group_name : "/ecs/${var.service_name}"
+}
+
+resource "aws_cloudwatch_log_group" "ecs" {
+  name              = local.effective_log_group_name
+  retention_in_days = 7
+}
+
+resource "aws_ecs_cluster" "this" {
+  name = var.cluster_name
+}
+
+resource "aws_iam_role" "ecs_task_execution_role" {
+  name               = "${var.service_name}-ecs-task-execution"
+  assume_role_policy = data.aws_iam_policy_document.ecs_task_assume_role_policy.json
+}
+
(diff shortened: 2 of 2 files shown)


infra/modules/scheduler/main.tf
```diff
@@ -0,0 +1,114 @@
...
+    ]
+  })
+}
+
+resource "aws_ecs_task_definition" "app" {
+  family                   = var.service_name
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = "512"
+  memory                   = "4096"
```
carlobortolan, line 71: I don't think that the scheduler actually needs 4GB of RAM, right?

(I'd guess that 264 CPU and 512 memory should be enough)

infra/modules/scheduler/main.tf
```diff
@@ -0,0 +1,114 @@
...
+      essential = true
+      environment = [
+        { name = "REDIS_URL", value = var.redis_endpoint },
+      ]
+      secrets = [
+        {
+          name      = "APP_CONFIG_JSON"
+          valueFrom = var.secrets_arn
+        }
+      ]
```
carlobortolan, line 86: APP_CONFIG_JSON was originally meant for the backend's .env

Do we have any specific .env variables needed for the scheduler? If not, we can leave this out.

infra/main.tf
```diff
@@ -102,6 +102,19 @@ module "ecs_dev" {
...
+module "scheduler" {
+  source            = "./modules/scheduler"
+  service_name      = "mediamind-service-scheduler-dev"
+  cluster_name      = local.cluster_name
+  container_image   = module.ecr.repository_url
+  redis_endpoint    = module.redis.endpoint
+  subnet_ids        = data.aws_subnets.selected.ids
+  vpc_id            = data.aws_vpc.selected.id
+  secrets_arn       = local.dev_secrets_arn
+  s3_backend_bucket = module.s3.bucket
```
carlobortolan, line 114: The scheduler doesn't need access to S3

infra/main.tf
```diff
@@ -102,6 +102,19 @@ module "ecs_dev" {
...
 
+module "scheduler" {
+  source            = "./modules/scheduler"
+  service_name      = "mediamind-service-scheduler-dev"
+  cluster_name      = local.cluster_name
+  container_image   = module.ecr.repository_url
+  redis_endpoint    = module.redis.endpoint
+  subnet_ids        = data.aws_subnets.selected.ids
+  vpc_id            = data.aws_vpc.selected.id
+  secrets_arn       = local.dev_secrets_arn
```
carlobortolan, line 113: The scheduler doesn't need access to our backend's .env

infra/main.tf
```diff
@@ -102,6 +102,19 @@ module "ecs_dev" {
   alb_security_group_id = module.alb.alb_security_group_id
 }
 
+module "scheduler" {
+  source            = "./modules/scheduler"
+  service_name      = "mediamind-service-scheduler-dev"
+  cluster_name      = local.cluster_name
+  container_image   = module.ecr.repository_url
```
carlobortolan, line 109: This will deploy the backend's image instead.

We need a separate ECR (with a separate repository_url ) for the scheduler; e.g., add:

```tf
module "ecr_scheduler" {
  source = "./modules/ecr"
  name   = "ecr-mediamind-scheduler"
}
```

infra/main.tf
```diff
@@ -102,6 +102,19 @@ module "ecs_dev" {
   alb_security_group_id = module.alb.alb_security_group_id
 }
 
+module "scheduler" {
+  source            = "./modules/scheduler"
+  service_name      = "mediamind-service-scheduler-dev"
```
carlobortolan, line 107: I'd name it: `mediamind-scheduler` to keep it simple

infra/modules/scheduler/main.tf
```diff
@@ -0,0 +1,114 @@
...
+
+resource "aws_ecs_service" "app" {
+  name            = var.service_name
+  cluster         = aws_ecs_cluster.this.id
+  task_definition = aws_ecs_task_definition.app.arn
+  desired_count   = 1
+  launch_type     = "FARGATE"
+  network_configuration {
+    subnets          = var.subnet_ids
+    assign_public_ip = true
```
carlobortolan, line 111: If we don’t need public access, set to `false`

carlobortolan (changes requested):

carlobortolan:
One additional comment: also update our pipelines so that the scheduler image is built and pushed to its ECR, and then a re-deployment is triggered (similar to how it currently is done with the backend)

carlobortolan (dismissed):

carlobortolan:
Looks good to me :+1: 

image

carlobortolan (approved):

PR #491, opened 2025-07-14
303 chatbot provide reports to llm

ricayanzon:


.github/workflows/apply-migrations.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-24.04
     defaults:
       run:
-        working-directory: ./apps/api
+        working-directory: ./apps/backend
```
carlobortolan, line 22: Are you sure that backend is a good idea? Might want to consider something like `matrix` instead :eyes:

carlobortolan (approved):
Looks good to me :+1:  ... just one comment: please ping me on slack before merging this, so that I can quickly update our Vercel deployments accordingly (otherwise they won't work with the updated structure)

carlobortolan (approved):

PR #494, opened 2025-07-14
Pdf as attachment in email

AlejandroGuevara05:
Had to add a migration to connect email and reports, since the report generation and email sending processes are separated and in the database there's no connection between the tables. 

Notes:
* Left report_id in emails as nullable true for the moment, to avoid problems
* The attachment has currently the name "report.pdf", suggestions?


AlejandroGuevara05:
Yes! Tested it a few times :)

Screenshot 2025-07-14 at 1 45 36 p m

Screenshot 2025-07-14 at 1 46 32 p m

carlobortolan:
@AlejandroGuevara05 I thought it was decided after the midterm that the reports should not be sent as a PDF attachment, but rather as a link to a file? Has this changed now?

nikbauer720:
> @AlejandroGuevara05 I thought it was decided after the midterm that the reports should not be sent as a PDF attachment, but rather as a link to a file? Has this changed now?

Yes, they have internal issues with that so we introduced an boolean on organizational level for this

AlejandroGuevara05:
Done!

nikbauer720 (approved):

nikbauer720 (approved):

PR #495, opened 2025-07-14
Small scraping fix

AlejandroGuevara05:
Scroll to the bottom of the page to load the content before extracting html. This solves the "Enable JavaScript" problem with Wiwo

Opened as draft because I still have to test it in docker


AlejandroGuevara05:
Nevermind, is working now, without this

PR #517, opened 2025-07-15
Include validity check after scraping

nikbauer720:


huaihaochuan (dismissed):
Looks good, thanks!

huaihaochuan (approved):

davidoof3141 (commented):
Could you also use article.note to say that the validity check failed

PR #520, opened 2025-07-15
Support only summary article page, Adjust coloring keyword badges article page

neher42:
image

### DIFF
--- a/apps/frontend/app/pages/article/article.tsx
+++ b/apps/frontend/app/pages/article/article.tsx
@@ -23,6 +23,9 @@ export function ArticlePage({
   const localizedHeadline = getLocalizedContent(article.article.headline);
   const localizedText = getLocalizedContent(article.article.text);
 
+  const onlySummary =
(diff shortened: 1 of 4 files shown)


milljoniaer:
so, actually I meant this read alert from the side to be where your new alert is now.

Or what do you think?

neher42:
image

apps/frontend/app/pages/article/article.tsx
```diff
@@ -23,6 +23,9 @@ export function ArticlePage({
   const localizedHeadline = getLocalizedContent(article.article.headline);
   const localizedText = getLocalizedContent(article.article.text);
 
+  const onlySummary =
+    article.article.text["en"] === "" && article.article.text["de"] === "";
```
milljoniaer, line 27: can we make this more bulletproof please? because in the future we will receive null for text.

```suggestion
    !(article.article.text?.["en"] && article.article.text?.["de"]);
```

This should work if text is null, if de is null and if de = "", because an empty string is falsy in JS

PR #521, opened 2025-07-15
Fix language switching article

neher42:


milljoniaer (changes requested):
I generally like your fix, but this makes `getLocalizedContent` a hook, because it calls a hook itself.

Thus you can only call it in the top of a component, not in the html, not after exit conditions. 

This currently leads to e.g. the SP overview not rendering. So pls check each occurance of `getLocalizedContent`

PR #534, opened 2025-07-15
Retouch Crawler Stats

MagMagnita:
- Adjusted Table und UI to match Admin Settings Layout

image

Idea: Do we want to add a panel on the top with for example: Total attempted, then ratio of positives ratio of negatives, etc. as a chart with colors maybe?

### DIFF
--- a/apps/frontend/app/pages/crawler-stats/crawler-stats.tsx
+++ b/apps/frontend/app/pages/crawler-stats/crawler-stats.tsx
@@ -18,6 +18,7 @@ import { type ColumnDef } from "@tanstack/react-table";
 import type { Stat } from "types/model";
 import { DataTable } from "~/custom-components/admin-settings/data-table";
 import { format } from "date-fns";
+import { Card } from "~/components/ui/card";
 
 export const CrawlerStatsPage = () => {
   const { t } = useTranslation();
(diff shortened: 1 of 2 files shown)


MagMagnita:
Feel free to leave comments also on the statistics part. The subscriptions ow is for subscriptions that were never scraped successfully (so the real negatives)

LeonardoZambranoL:
Nice job, looking nice 👍 

One request: When the texts are too long, make it so that it only shows the first 4 or so lines and add an "expand" or "Show more" button. and ofc the corresponding "show less".

apps/frontend/app/pages/crawler-stats/crawler-stats.tsx
```diff
@@ -103,19 +140,47 @@ export const CrawlerStatsPage = () => {
...
+            {(failureRatio * 100).toFixed(0)}%)
+          </Badge>
+        </div>
+        <div className="flex gap-4 mb-2">
+          <p>Subscriptions:</p>
+          <Badge>Sum attempted {data?.total_count}</Badge>
+          <Badge className="bg-red-600">
+            Never successful: {zeroSuccessful} (
+            {(zeroSuccessRatio * 100).toFixed(0)}%)
+          </Badge>
```
milljoniaer, line 176: localization pls

apps/frontend/app/pages/crawler-stats/crawler-stats.tsx
```diff
@@ -103,19 +140,47 @@ export const CrawlerStatsPage = () => {
...
+          <p className="font-semibold">Statistics</p>
+        </div>
+        <div className="flex gap-4 mb-2">
+          <p>Crawlers:</p>
+          <Badge>Sum attempted: {totalAttempted}</Badge>
+          <Badge className="bg-green-600">
+            Sum successful: {totalSuccessful} ({(successRatio * 100).toFixed(0)}
+            %)
+          </Badge>
+          <Badge className="bg-red-600">
```
milljoniaer, line 165:
```suggestion
          <Badge className="bg-destructive">
```
to match the theme

apps/frontend/app/pages/crawler-stats/crawler-stats.tsx
```diff
@@ -103,19 +140,47 @@ export const CrawlerStatsPage = () => {
...
+          setEndDate={setEndDate}
+        />
+        <div className="flex items-center">
+          <ChartPie className="mr-2" />
+          <p className="font-semibold">Statistics</p>
+        </div>
+        <div className="flex gap-4 mb-2">
+          <p>Crawlers:</p>
+          <Badge>Sum attempted: {totalAttempted}</Badge>
+          <Badge className="bg-green-600">
```
milljoniaer, line 161:
```suggestion
          <Badge className="bg-emerald-600">
```

milljoniaer:
and can we put "Crawlers" & Subscription above the badges and then the two badge blocks next to each other?

apps/frontend/app/pages/crawler-stats/crawler-stats.tsx
MagMagnita: Uhh yes totaly forgot that

MagMagnita: Localized now

PR #553, opened 2025-07-16
feat: added rss and breaking news crawler to the scheduler

specialfish9:


apps/scheduler/.env.example
```diff
@@ -6,6 +6,8 @@ SCHEDULER_INTERVAL=30
 
 EMAIL_JOB_INTERVAL=1800
 PIPELINE_JOB_INTERVAL=43200
+RSS_JOB_INTERVAL=1800
+BREAKING_NEWS_INTERVAL=3600
```
ricayanzon, line 10: @carlobortolan fyi

ricayanzon (dismissed):
dope.

ricayanzon:
let's wait for @davidoof3141's go

apps/scheduler/.env.example
carlobortolan: @specialfish9:
This needs to be updated in https://github.com/jst-seminar-rostlab-tum/eutop-mediamind/blob/b0692b95487fc0d0e8f90f847c627e51c5f75d14/infra/modules/scheduler/main.tf#L55-L62

specialfish9:
> Could we remove the pipeline job for now?

Yes, done!

PR #559, opened 2025-07-16
docs: add comprehensive documentation for ArticleCleanupService

dancing-shark:
close #549


davidoof3141:
I would personally prefer to have a high level documentation of the code. So specific explanations of arguments or code snippets do not really offer value as this should already be explained in the code. I would keep it really brief.

huaihaochuan (approved):

PR #570, opened 2025-07-17
Improve search profile page

neher42:
a

### DIFF
--- a/apps/frontend/app/pages/search-profile/search-profile.tsx
+++ b/apps/frontend/app/pages/search-profile/search-profile.tsx
@@ -1,4 +1,4 @@
-import { Book, FileText, Search } from "lucide-react";
+import { Book, FileText, Search, ChevronDown, ChevronUp } from "lucide-react";
 import { useEffect, useState } from "react";
 import { Link, useParams } from "react-router";
 import { client, useQuery } from "types/api";
@@ -17,7 +17,7 @@ import Text from "~/custom-components/text";
 import { getLocalizedContent, getPercentage } from "~/lib/utils";
 import { useNavigate } from "react-router";
 import { SidebarFilter } from "./sidebar-filter";
-import { ScrollArea, ScrollBar } from "~/components/ui/scroll-area";
+import { ScrollArea } from "~/components/ui/scroll-area";
 import {
   ArticlesSkeleton,
   SearchProfileSkeleton,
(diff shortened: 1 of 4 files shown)


MagMagnita:
- Thanks for the topic accordion, thas nice!

- I liked the bigger icons on the side bar, but we can also use smaller ones

- I don't like the removed border around the sources and topics, because the shadcn scroll bar is so thin you dont really see that its a scrollable list anymore. Also looks weird for low amount of topics bc of the white space:
image

- If you give the article container a fixed height, could you maybe reduce the padding/marging at the bot of the page?
image

- Yes we changed the overall scroll in app.css thats why the page scroll didnt work anymore, good catch!
- Skeleton looks good, too!

milljoniaer:
I dont think that we should add a fixed height to the right column: @neher42 that is what I mend with two scroll bars in each other, I think this is weird.

Generally, we have a hard time with small screens with our application. But we also focus on a german corporate and thus focusing on "normal" desktop sizes is okay.

image

But maybe we can address some of the topics you mentioned in slack this way:

- give the left sidebar more width, right now its limited to 400px maybe we can use a percontage as width and set a min-width of e.g. 500px (maybe also a max-width to not make it weird on huge screens), then at least the date picker should be in one line.
- make an article card within the right column less high: image can be smaller, text size of summary can be smaller, less padding, topics can go below summary on right of image
- we could hide all topics by default and only show them in expanded state -> we save another line of height

@LeonardoZambranoL what do you think? I really think we should not set a fixed height on the right container

neher42:
Adjusted Datepicker to always fit in one line with reduced minimum width
image

Made articles extend to the length of the page again as before, while keeping the filter sidebar fixed height (would also extend to end of page before)
image

When used on a small screen, the entire screen is scrollable, allowing users to have more than just two articles in their view. This also avoids having 2 scrollbars in the small sidebar in that scenario.
image

apps/frontend/app/pages/search-profile/search-profile.tsx
```diff
@@ -234,88 +288,121 @@ export function SearchProfileOverview() {
...
-                                    )}
-                                  </CardTitle>
-                                  <p className="line-clamp-3">
-                                    {getLocalizedContent(
-                                      match.article.summary,
-                                      i18n,
-                                    )}
-                                  </p>
+              <div className="bg-card rounded-lg border shadow-sm">
+                <ScrollArea className="p-2 h-[max(810px,calc(100vh-23rem))]">
```
milljoniaer, line 292: I am not a huge fan of these hardcoded calculated heights, bc they are very hard to maintain... (e.g. when you change the header size and so on).

milljoniaer: But having both the option to make it scrollable and span the full height as well as setting a max-height on small screens is not easy as well.

So if you are confident, that this works, feel free to merge

milljoniaer (approved):

PR #574, opened 2025-07-18
Feat/414 pdf design

NicolasArteaga:
This PR improves article rendering with better structure, visuals, and embedded content.

Changes
	•	Moved key metadata (e.g. published_at, keywords, reading_time) to the top of the article.
	•	Images are now downloaded and embedded from image_url.
	•	Switched to more readable fonts with broader character support.
	•	Minor fixes: logger cleanup, removed search_profile, design tweaks (padding, button alignment).
	•	Added translations and fixed a list-parsing bug.


LeonardoZambranoL:
please fix the conflicts

NicolasArteaga:
[output_translated (1).pdf](https://github.com/user-attachments/files/21328119/output_translated.1.pdf)

apps/backend/app/services/pdf_service/full_articles_elements.py
```diff
@@ -27,10 +33,44 @@ def create_full_articles_elements(news_items, dimensions, translator, styles):
             )
         )
 
+        # Use calculate_reading_time to get word count
```
nikbauer720, line 36: Is this based on the whole pdf (with the appendix)? Maybe this should only be done for each article once and not twice as it doubles in the appendix. I think it is implemented that way but just to make sure

nikbauer720 (commented):
Looks good. Just one question

apps/backend/crawl_stats_report.pdf
AlejandroGuevara05, line 1: Everything looks great! Just remove this pdf and we are good to go

AlejandroGuevara05: I'm assuming we don't really need it right?

AlejandroGuevara05 (approved):

PR #576, opened 2025-07-18
Fixes in translation

AlejandroGuevara05:
- Added the entities translation call to the pipeline
- Added the translated entities to the endpoint for the frontend
- Fixed an issue where langdetect crashes the pipeline when it received dates for language detection
- Added a threshold for translation, since in some cases lang detect is not good for short texts

Additionally, is not a translation problem but fixed not showing the article content for users not subscribed to the newspaper

Note: for the entities, frontend is now receiving a list of dicts instead of strings. They should probably update that and merge their PR and this one at the same time.


AlejandroGuevara05:
Added the translations of the new texts in the pdf to this PR, to fix the problem mentioned today morning

AlejandroGuevara05:
Converted to draft, frontend will do the respective updates in this same PR

nikbauer720 (commented):
Backend part is approved, just waiting for the frontend

AlejandroGuevara05:
Backend update: removed the use of the .mo file for the translations since it has given many problems. We are using a simple dictionary now

milljoniaer (approved):
Thx frontend part looks good

PR #577, opened 2025-07-19
Fix/localization

MagMagnita:
Fixed all issues regarding localization from slack chat


apps/frontend/app/custom-components/profile/edit/data-table-subscriptions.tsx
```diff
@@ -61,8 +61,8 @@ const getColumns = (
   {
     id: "data",
     accessorFn: (row) => row.data.name,
-    cell: ({ row }) => <div className="lowercase">{row.getValue("data")}</div>,
```
milljoniaer, line 64: 🥳

milljoniaer (approved):

PR #583, opened 2025-07-19
Deps: Remove unused, sort, and split into dev and prod deps

ricayanzon:


specialfish9 (approved):

davidoof3141:
There is actually a lot of stuff that is needed for selenium and playwright which is not referenced in the code.
Some dependencies also need to be installed explicitely through the requirements.txt for some reason.
These are needed for sure.
playwright
lxml_html_clean
fake-useragent
 I can run the stuff tomorrow in the afternoon to check more thoroughly.
@huaihaochuan Can you check which libraries you need for your custom crawlers?

ricayanzon:
@nikbauer720 we should check that everything works after this is merged

PR #593, opened 2025-07-20
Refactored Organization endpoints

nikbauer720:


apps/backend/app/repositories/organization_repository.py
```diff
@@ -177,8 +145,37 @@ async def get_pdf_as_link_by_recipient(recipient_email: str) -> bool:
                     )
                 )
             )
-            result = await session.execute(query)
-            pdf_as_link = result.scalar()
-            if pdf_as_link is None:
-                return True
-            return pdf_as_link
+            value = result.scalar()
+            return value if value is not None else True
```
ricayanzon, line 149:
```suggestion
            link = result.scalar()
            return link if link is not None else True
```

the naming was better before 👀 
why are you returining `True` if value is `None`? also - this is why it's better to always check for explicit conditions: if somewhere higher in the call stack of this function, you check for example for `if value` or `if not value`, since you return `True` here, it would treat it as if everything worked just fine (which didn't, because you're returning `True` here when it was `None`). this can be annoying to debug

apps/backend/app/services/organization_service.py
```diff
@@ -20,175 +21,76 @@ class OrganizationService:
...
-                role_map = {u.id: u.role for u in create_request.users}
-                user_ids = list(role_map.keys())
-
-                # Fetch and update each user
-                users = await UserRepository.get_by_ids(user_ids, session)
-                for user in users:
-                    user.organization_id = organization.id
-                    user.role = role_map.get(user.id, UserRole.member)
-                    await UserRepository.update_organization(user, session)
+            if create_request.users:
```
ricayanzon, line 33: I know this might seem like a good idea, but I have seen so many bugs caused by this implicit conditions that I really learned to avoid them: `if len(create_request.users) > 0` - the only situation in which this condition evaluates to `True`, is if the list of users is larger than 0. `create_request.users` on the other hand, evaluates to true if users is `True`, or any other truthy value. so it really could be anything here (see the previous comment where you returned `True` if the repository function didn't find anything - this would pass this condition too)

ricayanzon:
nice refactoring! only have a few comments :)

PR #595, opened 2025-07-20
Fix/hopefully the last fix

davidoof3141:


apps/backend/app/services/article_matching_service.py
```diff
@@ -217,28 +217,46 @@ async def _persist_matches(
...
+                    f"and article {art_id}"
+                )
+                try:
+                    if await MatchRepository.match_exists(
+                        session, profile_id, art_id
+                    ):
+                        skipped_count += 1
+                        self.logger.debug(
+                            f"Skipping article {art_id} for "
+                            f"profile {profile_id} - match already exists"
```
nikbauer720, line 231: Do we need to log this? This might overflow the log files

apps/backend/app/services/article_matching_service.py
```diff
@@ -217,28 +217,46 @@ async def _persist_matches(
 
             for _, (art_id, topic_id, score) in enumerate(matches):
                 # Check if match already exists for this profile and article
-                if await MatchRepository.match_exists(
-                    session, profile_id, art_id
-                ):
-                    skipped_count += 1
+                self.logger.debug(
```
nikbauer720, line 220: really necessary?

apps/backend/app/services/pipeline.py
```diff
@@ -15,6 +16,8 @@
     run_scraper,
 )
 
+logging.getLogger("LiteLLM").setLevel(logging.WARNING)
```
nikbauer720, line 19: why?

nikbauer720 (commented):
Do you want to log as much as possible?

apps/backend/app/services/pipeline.py
davidoof3141: The litellm logger overflowed the logs for each single request

apps/backend/app/services/article_matching_service.py
davidoof3141: This was there before. The git comparison is just messed up

apps/backend/data/subscriptions.json
```diff
@@ -466,7 +466,7 @@
     "paywall": true,
     "is_active": true,
     "login_config": {
-      "iframe_cookies": "//iframe[@id='sp_message_iframe_1251936']",
+      "iframe_cookies": "//iframe[@id='sp_message_iframe_1251936']",
```
huaihaochuan, line 469: Sadly the ifame number will constantly change:-( But it seems the only solution for maintainers to change it often.

apps/api/tests/unit_tests/__pycache__/test_search_profiles.cpython-313-pytest-8.3.5.pyc
huaihaochuan, line 1: Nice to have these pycache files deleted!

nikbauer720 (approved):

PR #597, opened 2025-07-21
Set use limit to LLM approach

AlejandroGuevara05:
Now we attempt the llm login approach once every 7 days, per subscription


nikbauer720 (dismissed):

apps/backend/app/services/web_harvester/web_harvester_orchestrator.py
```diff
@@ -156,28 +157,36 @@ def run_selenium_code(
...
+                        f"retrying login"
+                    )
+                    login_success = _handle_login_if_needed(
+                        subscription_updated, scraper, driver, wait
+                    )
+                else:
+                    scraper.logger.warning(
+                        f"Could not update login config for subscription "
+                        f"{subscription.name} with LLM approach."
+                    )
```
davidoof3141, line 189: In my fix branch I used the basic logger instead of the scraper logger. Is it possible if we used it for that as well?

apps/backend/app/services/web_harvester/web_harvester_orchestrator.py
```diff
@@ -156,28 +157,36 @@ def run_selenium_code(
...
 
         if subscription.paywall and not login_success:
             scraper.logger.warning(
```
davidoof3141, line 192: Here as well?

PR #604, opened 2025-07-21
Fix/landingpage previews

MagMagnita:
- Deleted "show only updated form dashboard preview
- Made Email a jst component, translated with i18n and added toasts for hrefs

image


LeonardoZambranoL (approved):
👍

PR #609, opened 2025-07-21
Fix/hydration error

MagMagnita:
- fixed i18n config
- fixed table header hydration error


apps/frontend/app/custom-components/article/article-meta-data-table.tsx
```diff
@@ -18,8 +12,10 @@ export function ArticleMetaDataTable({ article }: ArticleMetaDataTableProps) {
 
   return (
     <div className="border border-gray-200 rounded-lg p-3">
+      <div className="p-2 font-semibold">
+        {t("article-page.meta_data_header")}
+      </div>
       <Table>
-        <TableHead>{t("article-page.meta_data_header")}</TableHead>
```
milljoniaer, line 22: Why this change here?

apps/frontend/app/i18n.ts
```diff
@@ -1,23 +1,23 @@
 import i18n from "i18next";
 import { initReactI18next } from "react-i18next";
-import Backend from "i18next-http-backend";
```
milljoniaer, line 3: Should we remove this from package.json?

milljoniaer (commented):
Nice thanks for the fix, you included some unrelated changes in the PR, right?

But it's fine, just please have a look at the two comments :)

apps/frontend/app/custom-components/article/article-meta-data-table.tsx
MagMagnita: This also gave a hydration error, because tablehead has to be in a table row and used with table cell then. To keet the layout as it was, I just put the header in a div (since its not needed to be part of the table)

apps/frontend/app/i18n.ts
MagMagnita: I think so yes, I will remove it

milljoniaer (approved):

PR #615, opened 2025-07-22
370 chatbot include conversation ids in the email subject after first response to user

### LINKED ISSUE #370: Chatbot: Include conversation IDs in the email subject after first response to user
Currently, email conversations are matched based on `user.email`, `chat.subject` and `report_id`. But, it would be better if we have the ID of the email conversations in the subject already, in a format easy to parse (think if putting it in between of `[` and `]`). So then we can match with the corresponding conversation using the conversation ID and the user email, instead of matching with the entire subject.

This would probably implicate that all emails we send out regarding reports and breaking news, create an entry in the `email_conversations` table. Then, the `email_conversation.id` for that new entry is taken and included in the subject of the email send out. Finally, if a user replies to the email with the `email_conversation_id` in the subject, that id is parsed out, the conversation is loaded to the chatbot, and either the breaking news or report associated with that `email_conversation_id` is loaded into the context of the chatbot.

ricayanzon:
Closes #370 and #291

### DIFF
--- a/apps/backend/app/models/breaking_news.py
+++ b/apps/backend/app/models/breaking_news.py
@@ -1,11 +1,11 @@
-from uuid import UUID, uuid4
+from uuid import uuid4
 
 from sqlmodel import Field, SQLModel
 
 
 class BreakingNews(SQLModel, table=False):
-    id: UUID = Field(
-        default_factory=uuid4,
+    id: str = Field(
+        default_factory=lambda: str(uuid4()),
         primary_key=True,
         index=True,
     )
--- a/apps/backend/app/models/breaking_news.py
+++ b/apps/backend/app/models/breaking_news.py
@@ -1,11 +1,11 @@
-import uuid
+from uuid import UUID, uuid4
 
 from sqlmodel import Field, SQLModel
 
 
 class BreakingNews(SQLModel, table=False):
-    id: str = Field(
-        default_factory=lambda: str(uuid.uuid4()),
+    id: UUID = Field(
+        default_factory=uuid4,
         primary_key=True,
         index=True,
     )
--- a/apps/backend/app/models/email_conversation.py
+++ b/apps/backend/app/models/email_conversation.py
@@ -1,6 +1,6 @@
-import uuid
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Optional
+from uuid import UUID, uuid4
 
 from sqlalchemy import TIMESTAMP, Column, func
 from sqlmodel import Field, Relationship, SQLModel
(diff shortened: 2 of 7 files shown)


specialfish9 (dismissed):

apps/backend/app/models/breaking_news.py
```diff
@@ -1,11 +1,11 @@
-import uuid
+from uuid import UUID, uuid4
 
 from sqlmodel import Field, SQLModel
 
 
 class BreakingNews(SQLModel, table=False):
-    id: str = Field(
-        default_factory=lambda: str(uuid.uuid4()),
+    id: UUID = Field(
```
carlobortolan, line 7: This seems to _break_ the breaking news:

request:

```sh
curl -X 'POST' \
  'http://localhost:8000/api/v1/jobs/breaking-news' \
  -H 'accept: application/json' \
  -d ''
```

response:

```json
{
  "status": "failed",
  "message": "Failed to trigger breaking news crawling: 1 validation error for BreakingNews\nid\n  Input should be a valid UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-fA-F-], found `n` at 2 [type=uuid_parsing, input_value='eng-10805492', input_type=str]\n    For further information visit https://errors.pydantic.dev/2.11/v/uuid_parsing"
}
```

console logs:

```sh
2025-07-22 18:02:23,264 - INFO - app.api.v1.endpoints.jobs_controller - Triggering breaking news crawling
INFO:app.api.v1.endpoints.jobs_controller:Triggering breaking news crawling
INFO:     127.0.0.1:54884 - "POST /api/v1/jobs/breaking-news HTTP/1.1" 200 OK
```

carlobortolan (changes requested):

carlobortolan (dismissed):
Unrelated to this PR, but maybe we should respond with something else than a 5xx error when the conversation id is not included:

```json
{
  "detail": "Failed to send chat response: Failed to generate chatbot email response for user_id=73aaf38e-c12f-4e85-a058-744cf9b11693: couldn't create UUID from subject=Hello world []; Failed to generate chatbot email response for user_id=73aaf38e-c12f-4e85-a058-744cf9b11693: failed to extract conversation_id from subject. Sending default response.."
}
```
e.g., maybe a 400?

ricayanzon:
> Unrelated to this PR, but maybe we should respond with something else than a 5xx error when the conversation id is not included:
> 
> ```json
> {
>   "detail": "Failed to send chat response: Failed to generate chatbot email response for user_id=73aaf38e-c12f-4e85-a058-744cf9b11693: couldn't create UUID from subject=Hello world []; Failed to generate chatbot email response for user_id=73aaf38e-c12f-4e85-a058-744cf9b11693: failed to extract conversation_id from subject. Sending default response.."
> }
> ```
> 
> e.g., maybe a 400?

totally agreed - but this is only email, right? I'll skip this for now since there is no real frontend to send the errors to

PR #624, opened 2025-07-23
feat: Refactor PDF service and add Exeption handling

NicolasArteaga:
**PDFService Refactor & Logging**

- Set logger name to PDF_SERVICE for clearer logs.
- Moved section helpers to proper staticmethods inside the class.
- Updated all internal calls to use the new staticmethod names.
- Added a log when the PDF parser is activated.
- Improved error handling and logging for each PDF section.


apps/backend/app/services/pdf_service/pdf_service.py
```diff
@@ -45,110 +71,138 @@ async def create_pdf(
...
-        if timeslot not in ["morning", "afternoon", "evening"]:
+        try:
             logger.info(
-                "Invalid timeslot. Must be one of: ['morning', 'afternoon',"
-                " 'evening']"
+                f"Starting PDF generation for profile={search_profile_id}, timeslot={timeslot}, language={language}"
             )
+            # Obtain translator for the specified language
+            if timeslot not in ["morning", "afternoon", "evening"]:
+                logger.info(
```
nikbauer720, line 80: that should be en error not an info

nikbauer720 (commented):
Just one small point

PR #626, opened 2025-07-23
fix: no static imports from public warning 

MagMagnita:
- moved locales folder with translation files from public to app


milljoniaer (approved):
Thanks for the cleanup

PR #654, opened 2025-07-26
Delete and secure endpoints

nikbauer720:
I deleted all unused/ test endpoints and secured a few which were not secured yet. 

Please also have a look that I didn't delete too much and if there are still unused ones. (@milljoniaer maybe you can also have a quick look)

@specialfish9 @carlobortolan please secure the scheduler endpoints.


carlobortolan:
> please secure the scheduler endpoints.

@nikbauer720 do you know if our backend's env var `SECRET_KEY` is currently used for anything? If not, I'd say we can use it to secure the `/v1/jobs` endpoint.

@specialfish9 I'll update the envs, `scheduler/app/config.py` and update our tf scripts. Would be great if you could take care of updating the jobs controller and `scheduler/app/job.py` accordingly

nikbauer720:
> > please secure the scheduler endpoints.
> 
> @nikbauer720 do you know if our backend's env var `SECRET_KEY` is currently used for anything? If not, I'd say we can use it to secure the `/v1/jobs` endpoint.
> 
> @specialfish9 I'll update the envs, `scheduler/app/config.py` and update our tf scripts. Would be great if you could take care of updating the jobs controller and `scheduler/app/job.py` accordingly

Seams like it wasn't used so far. But anyway, if you don't change it, it cannot break anything.

PR #660, opened 2025-07-28
Refactor search profile service/repository

AlejandroGuevara05:
From:
search_profiles_service.py
To:
search_profiles_service
- __init__.py
- crud.py
- keywords.py
- matches.py
- subscriptions.py
- utils.py

From:
search_profile_repository.py
To:
search_profile_repository
- __init__.py
- crud.py
- utils.py


apps/backend/app/repositories/search_profile_repository/crud.py
```diff
@@ -0,0 +1,173 @@
...
+        if offset is not None and limit is not None:
+            query = query.offset(offset).limit(limit)
+
+        result = await session.execute(query)
+        return result.scalars().all()
+
+
+async def get_search_profile_by_id(
+    search_profile_id: UUID,
+) -> SearchProfile | None:
```
nikbauer720, line 128: I think the return value should not only be a `SearchProfile` as it includes more. Maybe there is already a schema or else make one.

nikbauer720 (commented):
Looks very nice!! Thanks a lot 🥇  Were you able to test it?

AlejandroGuevara05:
Yes! I triggered the report generation and sent some emails, to confirm functions were accessible, I thought that would suffice, or should I run the whole pipeline?

apps/backend/app/repositories/search_profile_repository/crud.py
AlejandroGuevara05: This would be only to make it explicit? Because the returned SearchProfile has the extra info loaded

